PR #17, opened 2017-07-17
FixedDueDateSchedules

FrancesWebb:
CIRCSTORE-9

I didn't make any version number changes, so if everything else is in order that will need to be applied.


dcrossleyau:
Using raml-cop to do external validation. It gives warnings about the "schema/schedule.json" where it does not like "date" as the "type". Following an example from "mod-users/userdata.json" ...

```
   "properties": {
     "from": {
-      "type": "date"
+      "type": "string",
+      "format": "date"
     },
... etc.
```

Then it is concerned about the examples with dummy dates "0000-00-00". Should such schedules just be omitted?

FrancesWebb:
Thanks, this change looks reasonable. Automatic checks passed, but I'll want to see what, if any, difference it makes in the jaxrs.

Regarding the zero-filled dummy dates, per scenario #9 on https://issues.folio.org/browse/UIS-14, a brand new Fixed Due Date Schedule should start with blank schedules. In the vein, I thought it was important to be able to store schedules that aren't filled out. I realize that I had been thinking that each schedule's three fields should be required but didn't put that in the json. I suppose validating that the dates be valid but allowing that the schedule's fields are all optional could be a good way to go. Do you think that seems like a clean way to go about it?

FrancesWebb:
The raml module builder doesn't understand "format":"date", and I think I'm going to go with "pattern":"\\d{4}-\\d{2}-\\d{2}" instead. That produces:

        @JsonProperty("from")
        @Pattern(regexp = "\\d{4}-\\d{2}-\\d{2}")
        @NotNull
        private String from;

in the jaxr model.

skoczko:
@funkymalc is there a problem with Jenkins configuration?

PR #103, opened 2018-08-31
preparation for CIRCSTORE-71

marcjohnson-kint:
*Purpose*

To improve the loans API tests in order to make it easier to add the required behaviour for CIRCSTORE-71

*Approach*

* Reduce the noise in the tests by extracting common activities and expectations around those activities, for example, when creating a loan we expect the response to be the created status code

* Allow a loan request builder to be created from an existing representation, so that it can be used to make changes when replacing records


marcjohnson-kint:
This was superceded by the CIRCSTORE-71 changes that were based upon it

marcjohnson-kint:
@jeremythuff Thank you for reviewing this, if you have any feedback please let me know

PR #122, opened 2018-12-17
CIRC-178 Adding 'actionComment' field to loan schema

dmytrotur:
CIRC-178 Adding 'actionComment' field to loan schema
Changes required by https://github.com/folio-org/mod-circulation/pull/171

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "loan-storage",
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
@dmytrotur @kostyantyn-kh Typically, FOLIO uses issues raised within a module's JIRA project to manage changes to the module. This has become especially important since we started deferring updating the news till formal releases, as it is necessary to be able to identify all of the issues involved in a version. Is there a circulation storage issue for this change

PR #123, opened 2018-12-19
CIRCSTORE-87 Adding 'actionComment' field to loan schema

kostyantyn-kh:


kostyantyn-kh:
added test for 'actionComment' field

PR #128, opened 2019-01-24
CIRCSTORE-92: Patron notice policy - Get by id

opochepam:


src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
```diff
@@ -137,6 +138,38 @@ public void canGetAllPatronNoticePolicies() throws MalformedURLException, Interr
...
+
+    String id = createPatronNoticePolicy(firstPolicy).getJson().getString("id");
+
+    CompletableFuture<JsonResponse> getCompleted = new CompletableFuture<>();
+
+    client.get(patronNoticePoliciesStorageUrl("/" + id), StorageTestSuite.TENANT_ID, ResponseHandler.json(getCompleted));
+
+    JsonResponse response = getCompleted.get(5, TimeUnit.SECONDS);
+
+    MatcherAssert.assertThat(response.getStatusCode(), CoreMatchers.is(200));
```
marcjohnson-kint: I think that the readability of these asserts could be improved with static imports of the assertion and matcher methods.
```suggestion
    assertThat(response.getStatusCode(), is(200));
```

opochepam: Good idea. Will do it. Thanks.

marcjohnson-kint: Thanks

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
```diff
@@ -116,6 +119,51 @@ public void postPatronNoticePolicyStoragePatronNoticePolicies(
...
+      try {
+        String tenantId = okapiHeaders.get(RestVerticle.OKAPI_HEADER_TENANT);
+        PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
+
+        Criteria criteria = new Criteria()
+          .addField("'id'")
+          .setOperation(Criteria.OP_EQUAL)
+          .setValue(patronNoticePolicyId);
+
+        pgClient.get(PATRON_NOTICE_POLICY_TABLE, PatronNoticePolicy.class, new Criterion(criteria), false, get -> {
```
julianladisch: Use

    pgClient.getById(PATRON_NOTICE_POLICY_TABLE, patronNoticePolicyId, PatronNoticePolicy.class, get -> {

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
```diff
@@ -116,6 +119,51 @@ public void postPatronNoticePolicyStoragePatronNoticePolicies(
...
+              .respond200WithApplicationJson(get.result().getResults().get(0))));
+        });
+      } catch (Exception e) {
+        logger.error(e);
+        asyncResultHandler.handle(Future.succeededFuture(
+          PostPatronNoticePolicyStoragePatronNoticePoliciesResponse.respond500WithTextPlain(e)));
+      }
+    });
+  }
+
```
julianladisch: Use

    public void getPatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePolicyId(
      String patronNoticePolicyId,
      Map<String, String> okapiHeaders,
      Handler<AsyncResult<Response>> asyncResultHandler,
      Context vertxContext) {

      PgUtil.getById(PATRON_NOTICE_POLICY_TABLE, PatronNoticePolicy.class, patronNoticePolicyId,
          okapiHeaders, vertxContext,
          GetPatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePolicyIdResponse.class,
          asyncResultHandler);
    }

domain-models-runtime has PgUtil.getById since version 23.3.0.

opochepam: Thanks. It looks much better now.

julianladisch: Well done!

PR #129, opened 2019-01-30
CIRCSTORE-90 Add "Delete Patron Notice Policy" API Endpoint

OleksiiMaksymov:
**Purpose** 
Backend: Delete Patron Notice Policy API Endpoint

**Links**
[CIRCSTORE-90](https://issues.folio.org/browse/CIRCSTORE-90)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "1.0",
+      "version": "0.4",
       "handlers": [
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.4",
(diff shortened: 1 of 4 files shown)


julianladisch:
The version number in the raml file is used for the documenation web pages: https://dev.folio.org/reference/api/
The version number in the module descriptor is used by okapi for dependency resolution.
The version numbers should match, and the raml version number should be adjusted to match the module descriptor version number.

These are the commits:
patron-notice-policy.raml v1.0: https://github.com/folio-org/mod-circulation-storage/commit/47c0f3bd990e93fabdf84de3bffa3557f076ec7a
module descriptor 1.0: https://github.com/folio-org/mod-circulation-storage/commit/448972338b9448954c80b9cb9be120de83dd90db
immediate module descriptor review fix 1.0 to 0.1: https://github.com/folio-org/mod-circulation-storage/commit/a0c46ba5019b2879d809f4c31c0f61c5aeee3643

All changes since then were compatible changes. The correct version number is 0.4. Both patron-notice-policy.raml and the module descriptor for patron-notice-policy-storage should have this version number.

The major version number is only changed if there is a breaking (non-compatible) change. This is not the case when the CRUD operations get added one by one. We do not increment the major version number at the time when all CRUD operations have been implemented.

PR #130, opened 2019-01-30
CIRCSTORE-96 Loan Policy Editor - Remove 'Alternate loan period for i…

kostyantyn-kh:
…tems with existing requests' field


marcjohnson-kint:
Given that this is a breaking change that requires compatibility updates from a variety of other modules, is it worth fixing some of the known breaking issues during this change?

For example, the disallowing of additional properties (which have been exploited a few times) and the stricter definition of periods?

@skoczko @zburke @mkuklis What do you think?

PR #144, opened 2019-02-27
Circstore 109 110 new request statuses

kostyantyn-kh:


kostyantyn-kh:
Thank you for your feedback
> For example, it is my understanding (from https://issues.folio.org/browse/CIRCSTORE-110) that only requests that haven’t started fulfilment should expire based upon the `request expiration date`.

Description defines two conditions for request expiration
 - Given a request with request status = "Open - Not yet filled"
 - When the request's "Request expiration date" passes
It's not clear for me what "haven’t started fulfilment" means? What schema field could i rely to find it out?

> Similarly (from https://issues.folio.org/browse/CIRCSTORE-109) that only requests that are waiting on the hold shelf should expire based upon the `hold shelf expiration date`.

As i understand it means "fulfilmentPreference"="Hold Shelf", am i correct?
if so will update query, thank you for pointing out.

> With the current implementation, what would happen if there was a request that was `Open - Not yet filled` yet the `hold shelf expiration date` had passed (but the `request expiration date` had not)?
No, if request has status `Open - Not yet filled` and no `request expiration date` - nothing would happen with this request, test scenarios cover this case.
Request expires when
 - status `Open - Not yet filled`
AND
 -  `request expiration date` had passed

The reason why request expiration and pickup expiration processes are combined in one place: the only difference between these two processes is initial query, where expired requests are selected

Again, thank you for useful feedback and please let me know if you have any comments or thoughts.

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -80,7 +55,8 @@ private static String getTenant(String nsTenant) {
     context.runOnContext(v -> {
       PostgresClient pgClient = PostgresClient.getInstance(vertx, tenant);
       pgClient.setIdField("_id");
-      String query = String.format("(status == \"%s\" AND requestExpirationDate < \"%s\") OR (status == \"%s\" AND holdShelfExpirationDate < \"%s\")",
+      String query = String.format(
+        String.format("%s OR %s", QUERY_TO_LOOK_FOR_EXPIRED_UNFILLED, QUERY_TO_LOOK_FOR_EXPIRED_AWAITING_PICKUP),
```
marcjohnson-kint: Thank you, that makes the distinction between the two more explicit

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -102,32 +78,17 @@ private static String getTenant(String nsTenant) {
...
   /* For each expired request set Closed status and remove position*/
   private static List<Future> closeRequests(Vertx vertx, Context context, String tenant, List<Request> requestList) {
     List<Future> futureList = new ArrayList<>();
     if (!requestList.isEmpty()) {
       for (Request request : requestList) {
-        if (request.getStatus() == OPEN_NOT_YET_FILLED) {
-          request.setStatus(CLOSED_UNFILLED);
-        } else {
-          request.setStatus(CLOSED_PICKUP_EXPIRED);
+        switch (request.getStatus()) {
```
marcjohnson-kint: I think Sonar is unhappy with this, and expects it be converted to two if statements

kostyantyn-kh: fixed

src/main/java/org/folio/support/ExpirationTool.java
kostyantyn-kh: thank you!

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -1521,6 +1522,584 @@ public void canDeleteARequest()
...
+
+    JsonResponse response = createCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(String.format("Failed to create request: %s", response.getBody()),
+      response.getStatusCode(), is(HttpURLConnection.HTTP_NO_CONTENT));
+
+    CompletableFuture<Void> getExpirationCF = new CompletableFuture<>();
+    ExpirationTool.doRequestExpiration(StorageTestSuite.getVertx(), StorageTestSuite.getVertx().getOrCreateContext()).setHandler(res -> {
+      getExpirationCF.complete(null);
+    });
+    getExpirationCF.get(5, TimeUnit.SECONDS);
```
marcjohnson-kint: I think the code for triggering expiration can be extracted to reduce duplication.

Something like
```suggestion
    expireRequests();

  private void expireRequests() throws InterruptedException, ExecutionException, TimeoutException {
    CompletableFuture<Void> expirationCompleted = new CompletableFuture<>();

    ExpirationTool.doRequestExpiration(StorageTestSuite.getVertx(),
      StorageTestSuite.getVertx().getOrCreateContext())
      .setHandler(res -> expirationCompleted.complete(null));

    expirationCompleted.get(15, TimeUnit.SECONDS);
  }
```

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -1521,6 +1522,584 @@ public void canDeleteARequest()
...
+      .hold()
+      .withId(id)
+      .withRequestExpiration(new DateTime(2017, 7, 30, 10, 22, 54, DateTimeZone.UTC))
+      .withItemId(itemId)
+      .withPosition(1)
+      .withStatus(OPEN_NOT_YET_FILLED)
+      .create();
+    CompletableFuture<JsonResponse> createCompleted = new CompletableFuture<>();
+    client.put(requestStorageUrl(String.format("/%s", id)),
+      requestRequest, StorageTestSuite.TENANT_ID,
```
marcjohnson-kint: I think most of the creation logic can be replaced with the `createRequest` method.

```suggestion
      createRequest(requestRequest);
```

kostyantyn-kh: replaced with createEntity (moved createRequest to superclass and renamed to createEntity to make it reusable)

src/test/java/org/folio/rest/api/RequestsApiTest.java
kostyantyn-kh: moved to separate method, thank you

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -0,0 +1,218 @@
...
+    pgClient.select(tenantQuery, future.completer());
+
+    return future.compose(rs -> CompositeFuture.all(rs.getRows()
+      .stream()
+      .map(row -> doRequestExpirationForTenant(vertx, context, getTenant(row.getString("nspname"))))
+      .collect(Collectors.toList())))
+      .map(compositeFuture -> null);
+  }
+
+  public static Future<Integer> doRequestExpirationForTenant(Vertx vertx, Context context, String tenant) {
```
Anatolii-Starkov: Does it require to be public?

PR #145, opened 2019-02-27
Circstore 103 load data

adamdickmeiss:
Reference data and sample data loading CIRCSTORE-103 CIRCSTORE-104


adamdickmeiss:
The sample is in sync with mod-users sample.. I'll sync up with inventory sample as well.

PR #146, opened 2019-03-04
CIRCSTORE-111 Backend: User initiated event trigger for loan related …

kostyantyn-kh:
…notices within notice policy


Anatolii-Starkov:
Please add a json example to mod-circulation-storage/ramls/examples/patron-notice-policy.json

PR #150, opened 2019-03-11
FOLIO-1782 load staff slips ref data using RMB tenant loading

hjiebsco:
This is to load two reference staff slips: **Hold** and **Transit**. To be compatible with RMB tenant loading mechanism, the PUT method for staff slips is modified to return 404 if the given staff slip id does not exist. Currently it creates and return 201. This change makes it more consistent with how other modules handle PUT (update) vs POST (create). The alternative approach would be to relax RMB [TenantLoading.java](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java#L228) to accept 201 as a success response code. Note: this PR is created on top of another PR [145](https://github.com/folio-org/mod-circulation-storage/pull/145)


marcjohnson-kint:
> To be compatible with RMB tenant loading mechanism, the PUT method for staff slips is modified to return 404 if the given staff slip id does not exist. Currently it creates and return 201. This change makes it more consistent with how other modules handle PUT (update) vs POST (create).

Generally, I expect PUT to create a record at the specified location (my interpretation of RFC 7231), rather than respond with a 404.

If FOLIO has instead standardised on responding with a 404 (and it appears so from a scan of PgUtil in RAML Module Builder), then that needs documenting. However that is beyond the scope of this change.

IMO changing the intended status code for an endpoint could be interpreted as a breaking implementation change (it isn’t a interface change because we allow 404 for PUT). I do think it’s unlikely clients will be affected by this, as they likely only use PUT for updates.

(Some endpoints, based upon the precursor to PgUtil, the StorageHelper, appear to do an upset, and respond with 204 on either creation or update, I imagine that is handled by 
https://github.com/folio-org/mod-inventory-storage/blob/bc7f1ad781c33ba5ddc3ff581f6778e600974b24/src/main/java/org/folio/rest/impl/StorageHelper.java#L395)

Just my ramblings, probably mostly out of scope for this pull request.

hjiebsco:
Looks like that @adamdickmeiss just updated RMB TenantLoading.java to accept PUT to return 201 and cut a new RMB release v23.10.0. In this case, I will update the code accordingly and bump up RMB dependency to v23.10.0 in mod-circulation-storage.

adamdickmeiss:
Yes.. I agree that PUT might create an item with the id given. But it doesn't in all cases, so that's what we have to deal with. If it does: fine.

PR #151, opened 2019-03-12
CIRCSTORE-115: Improve loan related query performance

hjiebsco:
Added a few indexes so loan related queries can be improved significantly (100x).


skoczko:
@marc-johnson Agree. @hjiebsco for issues like this where the fix is to add or tune the index, can we make sure we have a perf test for the API in Jenkin's "folio-perf-platform" and use the performance results for the API to confirm and close the issue?

hjiebsco:
Yes, @skoczko, those tuning are actually based on the SQLs generated by nightly folio-perf-platform tests.

PR #157, opened 2019-04-25
[CIRCSTORE-119] Not allow user to delete patron notice policy if currently in use

opochepam:
## Purpose
Resolves: CIRCSTORE-119
Not allow user to delete patron notice policy if currently in use
## Links
https://issues.folio.org/browse/CIRCSTORE-119

### DIFF
--- a/ramls/patron-notice-policy.raml
+++ b/ramls/patron-notice-policy.raml
@@ -82,11 +82,11 @@ traits:
         responses:
           204:
             description: "Source record deleted"
-          422:
-            description: "Unprocessable entity"
+          400:
+            description: "Bad request, e.g. malformed request body, query parameter or constraint violation."
             body:
-              application/json:
-                type: errors
+              text/plain:
+                example: "Cannot delete in use notice policy"
           404:
             description: "There is no source record for that patronNoticePolicyId"
             body:
--- a/ramls/patron-notice-policy.raml
+++ b/ramls/patron-notice-policy.raml
@@ -82,6 +82,11 @@ traits:
         responses:
           204:
             description: "Source record deleted"
+          422:
+            description: "Unprocessable entity"
+            body:
+              application/json:
+                type: errors
           404:
             description: "There is no source record for that patronNoticePolicyId"
             body:
--- a/ramls/patron-notice-policy.raml
+++ b/ramls/patron-notice-policy.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Patron Notice Policies
-version: v0.7
+version: v0.8
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost:9130
 
--- a/src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
@@ -163,8 +163,8 @@ public class PatronNoticePoliciesAPI implements PatronNoticePolicyStorage {
 
     findCirculationRules(pgClient)
       .map(CirculationRules::getRulesAsText)
-      .map(text -> text.indexOf(patronNoticePolicyId))
-      .compose(index -> index == -1 ? succeededFuture() : failedFuture(new NoticePolicyInUseException(IN_USE_POLICY_ERROR_MESSAGE)))
+      .map(text -> text.contains(patronNoticePolicyId))
+      .compose(contains -> contains ? failedFuture(new NoticePolicyInUseException(IN_USE_POLICY_ERROR_MESSAGE)) : succeededFuture())
       .compose(v -> deleteNoticePolicyById(pgClient, patronNoticePolicyId))
       .map(v -> DeletePatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePolicyIdResponse.respond204())
       .map(Response.class::cast)
--- a/src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
@@ -1,29 +1,49 @@
 package org.folio.rest.impl;
 
+import static io.vertx.core.Future.failedFuture;
+import static io.vertx.core.Future.succeededFuture;
+import static java.lang.String.format;
+import static java.util.Collections.singletonList;
+import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+
+import static org.folio.rest.impl.CirculationRulesAPI.CIRCULATION_RULES_TABLE;
+import static org.folio.rest.persist.Criteria.Criteria.OP_EQUAL;
+
+import java.util.Map;
+import java.util.UUID;
+
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.core.Response;
+
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.sql.UpdateResult;
+
+import org.z3950.zing.cql.cql2pgjson.CQL2PgJSON;
+
 import org.folio.rest.RestVerticle;
+import org.folio.rest.jaxrs.model.CirculationRules;
 import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.PatronNoticePolicies;
 import org.folio.rest.jaxrs.model.PatronNoticePolicy;
 import org.folio.rest.jaxrs.resource.PatronNoticePolicyStorage;
-import org.folio.rest.persist.Criteria.Limit;
-import org.folio.rest.persist.Criteria.Offset;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.persist.Criteria.Criteria;
+import org.folio.rest.persist.Criteria.Criterion;
+import org.folio.rest.persist.Criteria.Limit;
+import org.folio.rest.persist.Criteria.Offset;
 import org.folio.rest.persist.cql.CQLWrapper;
+import org.folio.rest.persist.interfaces.Results;
 import org.folio.rest.tools.utils.ValidationHelper;
-import org.z3950.zing.cql.cql2pgjson.CQL2PgJSON;
-
-import javax.ws.rs.core.Response;
-import java.util.Collections;
-import java.util.Map;
-import java.util.UUID;
+import org.folio.support.exception.NoticePolicyInUseException;
 
 public class PatronNoticePoliciesAPI implements PatronNoticePolicyStorage {
 
@@ -32,6 +52,8 @@ public class PatronNoticePoliciesAPI implements PatronNoticePolicyStorage {
   public static final String PATRON_NOTICE_POLICY_TABLE = "patron_notice_policy";
   public static final String STATUS_CODE_DUPLICATE_NAME = "duplicate.name";
   public static final String NOT_FOUND = "Not found";
+  private static final String INTERNAL_SERVER_ERROR = "Internal server error";
+  public static final String IN_USE_POLICY_ERROR_MESSAGE = "Cannot delete in use notice policy";
 
   @Override
   public void getPatronNoticePolicyStoragePatronNoticePolicies(
@@ -57,7 +79,7 @@ public class PatronNoticePoliciesAPI implements PatronNoticePolicyStorage {
         pgClient.get(PATRON_NOTICE_POLICY_TABLE, PatronNoticePolicy.class, fieldList, cql, true, false, get -> {
           if (get.failed()) {
             logger.error(get.cause());
-            asyncResultHandler.handle(Future.succeededFuture(
+            asyncResultHandler.handle(succeededFuture(
               GetPatronNoticePolicyStoragePatronNoticePoliciesResponse.respond500WithTextPlain(get.cause())));
             return;
           }
@@ -66,14 +88,14 @@ public class PatronNoticePoliciesAPI implements PatronNoticePolicyStorage {
             .withPatronNoticePolicies(get.result().getResults())
             .withTotalRecords(get.result().getResultInfo().getTotalRecords());
 
-          asyncResultHandler.handle(Future.succeededFuture(
+          asyncResultHandler.handle(succeededFuture(
             GetPatronNoticePolicyStoragePatronNoticePoliciesResponse
               .respond200WithApplicationJson(patronNoticePolicies)));
 
         });
       } catch (Exception e) {
         logger.error(e);
-        asyncResultHandler.handle(Future.succeededFuture(
+        asyncResultHandler.handle(succeededFuture(
           GetPatronNoticePolicyStoragePatronNoticePoliciesResponse.respond500WithTextPlain(e)));
       }
     });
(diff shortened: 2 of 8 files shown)


src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
```diff
@@ -209,12 +220,30 @@ public void cannotDeleteNonexistentPatronNoticePolicyId() throws MalformedURLExc
     assertThat(response.getBody(), is(NOT_FOUND));
   }
 
+  @Test
+  public void cannotDeleteInUsePatronNoticePolicy() throws InterruptedException, MalformedURLException,
+    TimeoutException, ExecutionException {
+
+    CirculationRules circulationRules = new CirculationRules()
+      .withRulesAsText(IN_USE_NOTICE_POLICY_ID);
```
marcjohnson-kint: This test seems to define the rules as just being the ID of the notice policy.

To me, this does not demonstrate an example of real usage, where the ID would be included as part of a set of rules.

opochepam: > This test seems to define the rules as just being the ID of the notice policy.
> 
> To me, this does not demonstrate an example of real usage, where the ID would be included as part of a set of rules.

replaced by real `rulesAsText` record

ramls/patron-notice-policy.raml
```diff
@@ -82,6 +82,11 @@ traits:
         responses:
           204:
             description: "Source record deleted"
+          422:
```
marcjohnson-kint: How was it decided that the response would be a 422/JSON rather than a 400/plain text (as is [already defined for deletion](https://github.com/folio-org/raml/blob/69f6074f5f48fd58e379e9ae76c892aa68c115ea/rtypes/item-collection.raml#L41))

As this is a new status code, is the client prepared to handle a 422 response?

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
```diff
@@ -209,12 +218,37 @@ public void cannotDeleteNonexistentPatronNoticePolicyId() throws MalformedURLExc
     assertThat(response.getBody(), is(NOT_FOUND));
   }
 
+  @Test
+  public void cannotDeleteInUsePatronNoticePolicy() throws InterruptedException, MalformedURLException,
+    TimeoutException, ExecutionException {
+
+    String inUsePolicyId = "16b88363-0d93-464a-967a-ad5ad0f9187c";
```
marcjohnson-kint: Why is this hard coded, rather than generated during the test?

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
```diff
@@ -209,12 +218,37 @@ public void cannotDeleteNonexistentPatronNoticePolicyId() throws MalformedURLExc
...
+
+    CompletableFuture<JsonResponse> putCompleted = new CompletableFuture<>();
+    client.put(rulesStorageUrl(), circulationRules, TENANT_ID, ResponseHandler.json(putCompleted));
+    putCompleted.join();
+
+    JsonResponse response = deletePatronNoticePolicy(inUsePolicyId);
+    String message = response.getJson().getJsonArray("errors").getJsonObject(0).getString("message");
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(message, is(IN_USE_POLICY_ERROR_MESSAGE));
```
marcjohnson-kint: This constant is defined in the implementation. Personally, I think tests should avoid coupling to internal implementation details where possible.

Using the same constant only checks that the constant is used, not that the message is what was intended.

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
opochepam: I prefer to use hard coded values rather then generated in tests, because it makes test debugging easier. Do you think such approach has some disadvantages?

ramls/patron-notice-policy.raml
opochepam: Actually client is not prepared yet. As 400 is already defined for constraint violation during deletion I will change status to 400. Thank you for the link.

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
```diff
@@ -209,12 +218,37 @@ public void cannotDeleteNonexistentPatronNoticePolicyId() throws MalformedURLExc
...
+      "fallback-policy: l 43198de5-f56a-4a53-a0bd-5a324418967a r 4c6e1fb0-2ef1-4666-bd15-f9190ff89060 " +
+      "n 122b3d2b-4788-4f1e-9117-56daa91cb75c m 1a54b431-2e4f-452d-9cae-9cee66c9a892: " +
+      "l d9cd0bed-1b49-4b5e-a7bd-064b8d177231 r 334e5a9e-94f9-4673-8d1d-ab552863886b n " + inUsePolicyId;
+
+    CirculationRules circulationRules = new CirculationRules()
+      .withRulesAsText(rulesAsText);
+
+    CompletableFuture<JsonResponse> putCompleted = new CompletableFuture<>();
+    client.put(rulesStorageUrl(), circulationRules, TENANT_ID, ResponseHandler.json(putCompleted));
+    putCompleted.join();
```
marcjohnson-kint: Join blocks until the future completes. If the future never completes, the test will be stuck on this line and never complete.

```suggestion
    putCompleted..get(5, TimeUnit.SECONDS);
```

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
marcjohnson-kint: I think it is a less representative example of actual usage and I think it makes it more likely that the test will pass/fail for the wrong reason (e.g. this test could be made to pass with a special case for this ID)

Others have also remarked that they feel hard coded IDs are easier for debugging. I think that is a signal that either the test or the system aren't providing sufficient diagnostic information (and there are likely plenty of cases of this). 

These are only my thoughts, I don't think this isn't something important enough to block merging.

opochepam:
> This is a breaking behavioural change. The implementation version needs to be updated to 8.0.0

Actually the version is 0.7 now. Should it be updated to 1.0?

ramls/patron-notice-policy.raml
opochepam: done

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
opochepam: done

src/test/java/org/folio/rest/api/PatronNoticePoliciesApiTest.java
opochepam: done

marcjohnson-kint:
> > This is a breaking behavioural change. The implementation version needs to be updated to 8.0.0
> 
> Actually the version is 0.7 now. Should it be updated to 1.0?

Apologies I may not have explained myself very well.

The version I am referring to is the module implementation version (currently 7.1.0-SNAPSHOT in the pom.xml).

Previously, it was possible to delete any patron notice policy. With this change, that is no longer possible. Which, in my opinion, makes this a breaking implementation / behavioural change.

An interface version change would also be needed for the addition of the 422/JSON response, however if that is no longer being done, it is not necessary to change the interface version.

Does that make sense?

opochepam:
> > > This is a breaking behavioural change. The implementation version needs to be updated to 8.0.0
> > 
> > 
> > Actually the version is 0.7 now. Should it be updated to 1.0?
> 
> Apologies I may not have explained myself very well.
> 
> The version I am referring to is the module implementation version (currently 7.1.0-SNAPSHOT in the pom.xml).
> 
> Previously, it was possible to delete any patron notice policy. With this change, that is no longer possible. Which, in my opinion, makes this a breaking implementation / behavioural change.
> 
> An interface version change would also be needed for the addition of the 422/JSON response, however if that is no longer being done, it is not necessary to change the interface version.
> 
> Does that make sense?

400/text response also was not defined in the raml so I suppose interface version anyway should be updated. And should it be updated to 1.0?

I am not sure about module implementation version. As I know currently module version is updated only during release procedure. Please correct me if I am wrong.

marcjohnson-kint:
> > > > This is a breaking behavioural change. The implementation version needs to be updated to 8.0.0
> > > 
> > > 
> > > Actually the version is 0.7 now. Should it be updated to 1.0?
> > 
> > 
> > Apologies I may not have explained myself very well.
> > The version I am referring to is the module implementation version (currently 7.1.0-SNAPSHOT in the pom.xml).
> > Previously, it was possible to delete any patron notice policy. With this change, that is no longer possible. Which, in my opinion, makes this a breaking implementation / behavioural change.
> > An interface version change would also be needed for the addition of the 422/JSON response, however if that is no longer being done, it is not necessary to change the interface version.
> > Does that make sense?
> 
> 400/text response also was not defined in the raml so I suppose interface version anyway should be updated. And should it be updated to 1.0?

Yes, the interface version does need updating too. I had thought the standard resource types were being used, my mistake.

Personally, I would go to 0.8 rather than 1.0, unless we are confident that the interface is stable enough that no breaking compatibility changes are coming with the upcoming patron notice work.

> I am not sure about module implementation version. As I know currently module version is updated only during release procedure. Please correct me if I am wrong.

There is some [debate about this](https://issues.folio.org/browse/FOLIO-1746). 

Personally, I think the version should be changed as soon as the behavioural compatibility is broken. 

Otherwise, we publish versions of 7.1.0 that aren't compatible with previous versions. Whilst this is ok, because it is a snapshot version. 

I think it effectively makes the number meaningless for diagnostic purposes. 

It also means the person responsible for releasing has to remember to check all of the outstanding changes to determine if any of them were breaking, and that is easy to miss.

To change the version pre-emptively would require releases to be planned in such a way that we would move to 8.0.0-SNAPSHOT on the assumption that a future change would be breaking (and this would contradict the release guidance about moving to the next minor snapshot version on release).

Does that make sense?

opochepam:
> > > > > This is a breaking behavioural change. The implementation version needs to be updated to 8.0.0
> > > > 
> > > > 
> > > > Actually the version is 0.7 now. Should it be updated to 1.0?
> > > 
> > > 
> > > Apologies I may not have explained myself very well.
> > > The version I am referring to is the module implementation version (currently 7.1.0-SNAPSHOT in the pom.xml).
> > > Previously, it was possible to delete any patron notice policy. With this change, that is no longer possible. Which, in my opinion, makes this a breaking implementation / behavioural change.
> > > An interface version change would also be needed for the addition of the 422/JSON response, however if that is no longer being done, it is not necessary to change the interface version.
> > > Does that make sense?
> > 
> > 
> > 400/text response also was not defined in the raml so I suppose interface version anyway should be updated. And should it be updated to 1.0?
> 
> Yes, the interface version does need updating too. I had thought the standard resource types were being used, my mistake.
> 
> Personally, I would go to 0.8 rather than 1.0, unless we are confident that the interface is stable enough that no breaking compatibility changes are coming with the upcoming patron notice work.
> 
> > I am not sure about module implementation version. As I know currently module version is updated only during release procedure. Please correct me if I am wrong.
> 
> There is some [debate about this](https://issues.folio.org/browse/FOLIO-1746).
> 
> Personally, I think the version should be changed as soon as the behavioural compatibility is broken.
> 
> Otherwise, we publish versions of 7.1.0 that aren't compatible with previous versions. Whilst this is ok, because it is a snapshot version.
> 
> I think it effectively makes the number meaningless for diagnostic purposes.
> 
> It also means the person responsible for releasing has to remember to check all of the outstanding changes to determine if any of them were breaking, and that is easy to miss.
> 
> To change the version pre-emptively would require releases to be planned in such a way that we would move to 8.0.0-SNAPSHOT on the assumption that a future change would be breaking (and this would contradict the release guidance about moving to the next minor snapshot version on release).
> 
> Does that make sense?

Yes, it makes sense to me. Thank you for the explanation.

Committed changes with updated module implementation and interface versions.

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
```diff
@@ -159,38 +188,84 @@ public void putPatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePolicy
...
+    Future<UpdateResult> future = Future.future();
+    pgClient.delete(PATRON_NOTICE_POLICY_TABLE, new Criterion(criteria), future.completer());
+
+    return future.map(UpdateResult::getUpdated)
+      .compose(updated -> updated > 0 ? succeededFuture() : failedFuture(new NotFoundException(NOT_FOUND)));
+  }
+
+  private Response mapExceptionToResponse(Throwable t) {
+
+    if (t.getClass() == NotFoundException.class) {
```
marcjohnson-kint: Is there a reason why this is does a reference comparison, rather than using `instanceOf`?

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
```diff
@@ -137,9 +158,17 @@ public void deletePatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePol
...
+    PostgresClient pgClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+
+    findCirculationRules(pgClient)
+      .map(CirculationRules::getRulesAsText)
+      .map(text -> text.contains(patronNoticePolicyId))
+      .compose(contains -> contains ? failedFuture(new NoticePolicyInUseException(IN_USE_POLICY_ERROR_MESSAGE)) : succeededFuture())
+      .compose(v -> deleteNoticePolicyById(pgClient, patronNoticePolicyId))
+      .map(v -> DeletePatronNoticePolicyStoragePatronNoticePoliciesByPatronNoticePolicyIdResponse.respond204())
+      .map(Response.class::cast)
+      .otherwise(this::mapExceptionToResponse)
```
marcjohnson-kint: I'm not a fan of managing flow of control (e.g. validation) with exceptions. However, I understand that we don't really have an alternative pattern in most modules

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
opochepam: I'm looking for matching an exact class therefore I don't see any reason to use `instanceOf` here.

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
opochepam: I had doubts about this but such approach seemed to me the most convenient here.

marcjohnson-kint: I think it is convenient, it seems to be the prevailing method in the storage modules

src/main/java/org/folio/rest/impl/PatronNoticePoliciesAPI.java
marcjohnson-kint: Ok (I don't follow why it's necessary to match the exact class, but that's ok).

This code will break if the exception is null, is it worth being defensive at the beginning of the method?

opochepam: If the exception is null, the method returns 500 response. Probably it is not a proper behavior but exception cannot be null here as the method is invoked only when the future is failed. So I think it does not make a lot of sense to handle null exception here.

marcjohnson-kint: Hmm, if the exception is null, I believe there is likely to be no response at all, because `t.getClass()` will throw a fresh `NullPointerException`.

I'm ok with an unknown cause resulting in a 500. I can't think of an alternative, given the information available?

Does a failed future guarantee that there is an exception present?

opochepam: Sorry, you're right about `NullPointerException`. 

Yes, as I far as I know a failed future guarantee that there is an exception present (if we fail a future without any exception, the future will be failed with `io.vertx.core.impl.NoStackTraceThrowable`).

opochepam: Currently `mapExceptionToResponse` method is used only to map exceptions for one endpoint. The method will be extending to handle more cases if we will want to use it for other endpoints.

marcjohnson-kint: Ok, if we are confident this won't be invoked with null, then we can resolve this

PR #162, opened 2019-05-20
[CIRCSTORE-126] Implement database transactions in the request expiration process

opochepam:
## Purpose
Resolves: CIRCSTORE-126
Not allow user to delete patron notice policy if currently in use
## Links
https://issues.folio.org/browse/CIRCSTORE-126

### DIFF
--- a/src/main/java/org/folio/support/ExpirationTool.java
+++ b/src/main/java/org/folio/support/ExpirationTool.java
@@ -1,6 +1,10 @@
 package org.folio.support;
 
+import static io.vertx.core.Future.succeededFuture;
 import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
 
 import static org.folio.rest.impl.RequestsAPI.REQUEST_TABLE;
 import static org.folio.rest.jaxrs.model.Request.Status.CLOSED_PICKUP_EXPIRED;
@@ -12,8 +16,10 @@ import static org.folio.rest.jaxrs.model.Request.Status.OPEN_NOT_YET_FILLED;
 import java.text.SimpleDateFormat;
 import java.time.ZoneOffset;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TimeZone;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
@@ -46,7 +52,7 @@ public class ExpirationTool {
     return future.compose(rs -> CompositeFuture.all(rs.getRows()
       .stream()
       .map(row -> doRequestExpirationForTenant(vertx, getTenant(row.getString("nspname"))))
-      .collect(Collectors.toList()))
+      .collect(toList()))
       .map(all -> null));
   }
 
@@ -58,7 +64,7 @@ public class ExpirationTool {
 
     pgClient.startTx(conn -> getExpiredRequests(conn, vertx, tenant)
       .compose(requests -> closeRequests(conn, vertx, tenant, requests))
-      .compose(v -> getOpenRequests(conn, vertx, tenant))
+      .compose(itemIds -> getOpenRequestsByItemIds(conn, vertx, tenant, itemIds))
       .compose(requests -> reorderRequests(conn, vertx, tenant, requests))
       .setHandler(v -> {
         if (v.failed()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -42,44 +46,67 @@ private ExpirationTool() {
     return future.compose(rs -> CompositeFuture.all(rs.getRows()
       .stream()
       .map(row -> doRequestExpirationForTenant(vertx, getTenant(row.getString("nspname"))))
-      .collect(Collectors.toList())));
+      .collect(Collectors.toList()))
+      .map(all -> null));
```
marcjohnson-kint: Can you help me understand this change? What collect the set of future for each tenant and then map them to null?

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -42,44 +46,67 @@ private ExpirationTool() {
...
       OPEN_NOT_YET_FILLED.value(),
       OPEN_AWAITING_PICKUP.value(),
       df.format(new Date()));
 
-    pgClient.get(REQUEST_TABLE, Request.class, "jsonb", where, false, false, false, future.completer());
 
-    return future.map(Results::getResults);
+    String fullTableName = format("%s.%s", PostgresClient.convertToPsqlStandard(tenant), REQUEST_TABLE);
+    String query = format("SELECT jsonb FROM %s %s", fullTableName, where);
+    pgClient.select(conn, query, future);
```
marcjohnson-kint: Is this because there isn’t a way to do a get whilst specifying a connection / transaction?

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -103,51 +137,61 @@ private static Request changeRequestStatus(Request request) {
...
 
-  private static Future<UpdateResult> resetPositionsForOpenRequests(Vertx vertx, String tenant) {
+  private static Future<UpdateResult> resetPositionsForOpenRequests(AsyncResult<SQLConnection> conn, Vertx vertx,
+                                                                    String tenant) {
 
-    String fullTableName = String.format("%s.%s", PostgresClient.convertToPsqlStandard(tenant), REQUEST_TABLE);
+    String fullTableName = format("%s.%s", PostgresClient.convertToPsqlStandard(tenant), REQUEST_TABLE);
 
-    String query = String.format("UPDATE %1$s SET jsonb = jsonb - 'position' WHERE " +
+    String sql = format("UPDATE %1$s SET jsonb = jsonb - 'position' WHERE " +
```
marcjohnson-kint: This seems to reset the positions for all open requests no matter if there was an expired request for that item.

Am I following this correctly?

src/main/java/org/folio/support/ExpirationTool.java
opochepam: I made this change to make method return type more obvious, `Future<Void>` instead of `Future<CompostiteFuture>`. It means that we expect a future without any result(`Void` result). As `CompositeFuture.all` is used here the method result will succeed if all composed results are succeeded.

src/main/java/org/folio/support/ExpirationTool.java
opochepam: Yes. There is no get method overloading with `where` parameter.

src/main/java/org/folio/support/ExpirationTool.java
opochepam: Yes, you are. Thank you for this point. I have already made corresponding changes.

src/main/java/org/folio/support/ExpirationTool.java
marcjohnson-kint: Ok, thanks for explaining it to me

src/main/java/org/folio/support/ExpirationTool.java
marcjohnson-kint: Is it worth raising a change for the RAML Module Builder for this to be added in future?

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -103,74 +154,103 @@ private static Request changeRequestStatus(Request request) {
...
+  private static Future<Void> updateRequestsPositions(AsyncResult<SQLConnection> conn, Vertx vertx,
+                                                      String tenant, List<Request> requests) {
 
     AtomicInteger pos = new AtomicInteger(1);
+    Future<Void> future = succeededFuture();
 
-    return CompositeFuture.all(requests.stream()
-      .map(request -> updateRequest(vertx, tenant, request.withPosition(pos.getAndIncrement())))
-      .collect(Collectors.toList()));
+    for (Request request : requests) {
```
marcjohnson-kint: How does this know what the correct position is? How are the requests order during this step?

(Apologies, I probably should've spotted this during the original work, it isn't really part of this change)

marcjohnson-kint: Or does this work because the affected requests are ordered by position and translating them to a map by item ID does not affect that ordering?

opochepam: At first requests which need reordering is grouped by itemId and than request reordering process is performed separately for each item.
Probably  I made a mistake here. Seems like requests list should be sorted by position before reordering.

opochepam: I've just made a change. Now request list is sorted by position before reordering and there is a guarantee that requests are in right order before setting new positions.

marcjohnson-kint: Thank you

src/main/java/org/folio/support/ExpirationTool.java
opochepam: Probably having such overloading would be better, but I am ok with such approach.

PR #164, opened 2019-05-22
[CIRCSTORE-127] Request update database trigger

opochepam:
## Purpose
Resolves: CIRCSTORE-127
When the request is updated from 'Open - Awaiting pickup' status to 'Closed - Pickup expired' or 'Closed - Cancelled' and the request is ont the top of a queue (has position 1) add new field to the request with transition timestamp (filed name - awaitingPickupRequestClosedDate).
## Approach
Request update database trigger
## Links
https://issues.folio.org/browse/CIRCSTORE-127


opochepam:
> I'll admit I was surprised that this has been implemented as a trigger. Please could this be documented as design notes in the readme.
> 
> It could also be useful for there be an API test to demonstrate the general behaviour (not all examples) from an API perspective.

Added Design notes to the readme file.
Added a couple of API test for request update.

ramls/request.json
```diff
@@ -170,6 +170,11 @@
       "type": "object",
       "description": "Tags",
       "$ref": "raml-util/schemas/tags.schema"
+    },
```
marcjohnson-kint: This new property requires a minor version increment in the interface version (both in the RAML and module descriptor)

opochepam: done

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -1500,6 +1505,66 @@ public void canDeleteARequest()
...
+    String requestId = createEntity(request, requestStorageUrl()).getJson().getString("id");
+    request.put("status", CLOSED_PICKUP_EXPIRED);
+
+    CompletableFuture<Response> putCompleted = new CompletableFuture<>();
+    client.put(requestStorageUrl("/" + requestId), request, TENANT_ID, ResponseHandler.empty(putCompleted));
+    putCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject updatedRequest = getById(requestStorageUrl("/" + requestId));
+
+    assertThat(updatedRequest.getString("awaitingPickupRequestClosedDate"), is(notNullValue()));
```
marcjohnson-kint: There are other tests that check whether a date property is similar date and time to now, using a custom matcher. 

This would provide more feedback than the property just being present.

See [this example](https://github.com/folio-org/mod-circulation/blob/666d15a77614d55ae66cde6ac07f050e3fba876f/src/test/java/api/loans/CheckInByBarcodeTests.java#L87) from mod-circulation

opochepam: Thank you for the example.
Reworked the test with a custom matcher.

src/test/java/org/folio/rest/api/RequestUpdateTriggerTest.java
```diff
@@ -62,21 +66,30 @@ void updateRequestTriggerTest(String oldStatus,
...
       .compose(v -> updateRequest(id, request.withStatus(fromValue(newStatus))))
       .compose(v -> getRequest(id))
       .setHandler(updatedRequest -> future.complete(updatedRequest.result()));
 
-    Request updatedRequest = future.get(5, TimeUnit.SECONDS);
+    JsonObject updatedRequest = future.get(5, TimeUnit.SECONDS);
+
 
-    assertThat(updatedRequest.getAwaitingPickupRequestClosedDate() != null, is(awaitingPickupRequestClosedDatePresent));
+    if (awaitingPickupRequestClosedDatePresent) {
```
marcjohnson-kint: Rather than introduce an if statement in the tests, an alternative is to assert for the presence of the property and then if the value is similar to expected.

If the first assert fails, the test will stop, there is no need for conditional execution.

opochepam: Right, but the property can be not presented and it is normal scenario. The property presence is a parameter of parameterized test defined in Csv source https://github.com/folio-org/mod-circulation-storage/pull/164/files#diff-aef83ff3f59a523d89c77c8b22f052c8R58

So if we are not expecting this property presented than we don't need to perform the check if the value is similar to expected.

marcjohnson-kint: If we are not expecting the property to be present, we should specifically check for that.

Oh, personally I would split the scenarios where the property is expected to be present and where it is not.

I find combining them together and using conditionals in the test makes it harder to understand the behaviour and increases the chances that the tests are brittle or unreliable.

opochepam: Agree that conditional in the test looks bad.

Initially I decided to use such approach (parameterized test with csv source) to keep all scenarios in one place. Where each row is test's parameters.

For example:
`"Open - Awaiting pickup | Closed - Pickup expired | true"` means that the property is present after status transition from `Open - Awaiting pickup` to `Closed - Pickup expired`.

I can get rid of `awaitingPickupRequestClosedDatePresent` test parameter  and split the test into two ones, the first for scenarios when the property must be present and the second one where not.

marcjohnson-kint: Thank you

opochepam: Split the test and added more cases for scenario when the property should not be presented.

julianladisch:
I'm also surprised by two things:
* The logic is implemented by a trigger. This is very unusual. There should be at least some comment in putRequestStorageRequestsByRequestId warning that there is a trigger that overwrites the awaitingPickupRequestClosedDate.
* The logic that updates awaitingPickupRequestClosedDate is business logic. It does not belong to mod-circulation-storage (storage) but to mod-circulation or to the front-end.

opochepam:
> I'm also surprised by two things:
> 
> * The logic is implemented by a trigger. This is very unusual. There should be at least some comment in putRequestStorageRequestsByRequestId warning that there is a trigger that overwrites the awaitingPickupRequestClosedDate.
> * The logic that updates awaitingPickupRequestClosedDate is business logic. It does not belong to mod-circulation-storage (storage) but to mod-circulation or to the front-end.

The `awaitingPickupRequestClosedDate` property have to be added to the request in two cases 

- request is updated from `Open - Awaiting pickup` status to `Closed - Pickup expired` (performed by request expiration job based on `PeriodicAPI` which is located in the `mod-circulation-storage`

- request is updated from `Open - Awaiting pickup` status to `Closed - Cancelled` (request is canceled manually)

Anyway part of the logic is in the `mod-circulation-storage` so it has been decided to implement the logic as trigger in order to keep the logic in one place.

Do you mean description in the raml file by a comment in putRequestStorageRequestsByRequestId?

marcjohnson-kint:
> * The logic is implemented by a trigger. This is very unusual. 

Agreed, this solution was a compromise proposed by @kostyantyn-kh as it was deemed too time consuming and complex to move the expiration behaviour to mod-circulation before the end of quarter 2.

> There should be at least some comment in putRequestStorageRequestsByRequestId warning that there is a trigger that overwrites the awaitingPickupRequestClosedDate.

Yes. I wonder if that property should be marked `readOnly` for the moment?

> * The logic that updates awaitingPickupRequestClosedDate is business logic. It does not belong to mod-circulation-storage (storage) but to mod-circulation or to the front-end.

This is another aspect of the compromise solution. The only part of the system that is aware of the state transition from `Open - Awaiting pickup` to `Closed - Pickup expired` is storage, which means that it needs to be implemented here.

It also means that the custom report being built in mod-circulation has to infer this transition from the presence of the `awaitingPickupRequestClosedDate` date on a request.

marcjohnson-kint:
> Do you mean description in the raml file by a comment in putRequestStorageRequestsByRequestId?

This shouldn't go in the RAML file as that documents the interface, and this implementation specific.

Marking the property as `readOnly` might be worthwhile to better communicate this.

opochepam:
> Marking the property as `readOnly` might be worthwhile to better communicate this.

Marked `awaitingPickupRequestClosedDate` property as `readonly`.

README.MD
```diff
@@ -139,3 +139,13 @@ The built artifacts for this module are available.
 See [configuration](https://dev.folio.org/download/artifacts) for repository access,
 and the [Docker image](https://hub.docker.com/r/folioorg/mod-circulation-storage/).
 
+## Design notes
+
+### Request status transition
```
marcjohnson-kint: Please can some context be added to this, to refer to the report in mod-circulation and the reason why this has been implemented here, rather than in mod-circulation

README.MD
```diff
@@ -139,3 +139,13 @@ The built artifacts for this module are available.
...
+
+There is database trigger for request update. It works in the following way:
+
+Given request with status 'Open - Awaiting pickup'.
+The request is updated and new status is 'Closed - Pickup expired' or 'Closed - Cancelled'.
```
marcjohnson-kint:
```suggestion
When the request is updated and new status is 'Closed - Pickup expired' or 'Closed - Cancelled'
```

README.MD
```diff
@@ -139,3 +139,13 @@ The built artifacts for this module are available.
 See [configuration](https://dev.folio.org/download/artifacts) for repository access,
 and the [Docker image](https://hub.docker.com/r/folioorg/mod-circulation-storage/).
 
+## Design notes
+
+### Request status transition
+
+There is database trigger for request update. It works in the following way:
+
+Given request with status 'Open - Awaiting pickup'.
```
marcjohnson-kint:
```suggestion
Given a request with status 'Open - Awaiting pickup'
```

README.MD
```diff
@@ -139,3 +139,13 @@ The built artifacts for this module are available.
...
+The request is updated and new status is 'Closed - Pickup expired' or 'Closed - Cancelled'.
+Than an update trigger adds new property 'awaitingPickupRequestClosedDate' to the request jsonb.
```
marcjohnson-kint:
```suggestion
Then an update trigger adds the property 'awaitingPickupRequestClosedDate' to the request jsonb
```

src/main/resources/templates/db_scripts/requestUpdateTrigger.sql
```diff
@@ -0,0 +1,20 @@
...
+  IF OLD.jsonb->>'status' = 'Open - Awaiting pickup' AND
+    NEW.jsonb->>'status' IN ('Closed - Pickup expired', 'Closed - Cancelled') THEN
+    NEW.jsonb = jsonb_set(NEW.jsonb, '{awaitingPickupRequestClosedDate}', to_jsonb(current_timestamp), true);
+  	END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE 'plpgsql';
+
+
+DROP TRIGGER IF EXISTS ${table.tableName}_update_trigger ON ${myuniversity}_${mymodule}.${table.tableName} CASCADE;
```
marcjohnson-kint: Why is there a need to cascade the removal of a trigger?

opochepam: Apologies, cascade is redundant here. I copied the example of trigger removal from one of the RMB module's script. Now cascade is removed.

PR #165, opened 2019-05-22
Improve rules and policies for testing CIRCSTORE-131

marcjohnson-kint:
Recently it has been mentioned that the default loan and request policies could be more useful for testing.

In order to make testing easier in the hosted environments, change the circulation rules and policies to
* allow any request types
* use rolling policy for most circumstances
* Add a recall management policy to the rolling policy

The change also removes the second example notice policy because it is the same as the first


reference-data/loan-policy-storage/loan-policies/rolling.json
```diff
@@ -6,8 +6,8 @@
   "loansPolicy" : {
     "profileId" : "Rolling",
     "period" : {
-      "duration" : 1,
-      "intervalId" : "Months"
+      "duration" : 60,
+      "intervalId" : "Days"
```
kostyantyn-kh: What is the purpose for this change?

marcjohnson-kint: Cate asked for it to be made when we talked about what policy would best help them test

PR #169, opened 2019-05-27
[CIRCSTORE-132] CRUD for scheduled tasks

opochepam:
## Purpose
Resolves: CIRCSTORE-132
Implement CRUD for scheduled tasks
## Links
https://issues.folio.org/browse/CIRCSTORE-132


descriptors/ModuleDescriptor-template.json
```diff
@@ -283,6 +283,47 @@
...
+          "pathPattern": "/scheduled-notice-storage/scheduled-notices/{id}",
+          "permissionsRequired": ["scheduled-notice-storage.scheduled-notices.item.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/scheduled-notice-storage/scheduled-notices",
+          "permissionsRequired": ["scheduled-notice-storage.scheduled-notices.item.post"]
+        },
+        {
+          "methods": ["POST"],
```
marcjohnson-kint: Is this endpoint for batch updating of records?

opochepam: Yes

descriptors/ModuleDescriptor-template.json
```diff
@@ -283,6 +283,47 @@
         }
       ]
     },
+    {
+      "id": "scheduled-notice-storage",
+      "version": "1.0",
```
marcjohnson-kint: I'd like to suggest that we mark this interface as experimental, by using `0.1`, given that it is likely to change

opochepam: Changed version to 0.1

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: This does not appear to be used by the [implementation of CIRC-322](https://github.com/folio-org/mod-circulation/pull/258/files), why has it been implemented in this first piece of work?

opochepam: Initially it was decided that CIRC-322 needs bulk save endpoint to save multiple scheduled notices in one request. But during implementation it was decided to use usual POST to save entities separately. Such decision was made because in most cases amount of notices not as big to require bulk save (about 1-3). So I think I should remove this endpoint for now as it is not used.

marcjohnson-kint: That's great, thank you. It gives us time to better establish the pattern for batch record creation or updating.

ramls/scheduled-notice-storage.raml
```diff
@@ -0,0 +1,53 @@
...
+  language: !include raml-util/traits/language.raml
+  pageable: !include raml-util/traits/pageable.raml
+  searchable: !include raml-util/traits/searchable.raml
+  validate: !include raml-util/traits/validation.raml
+
+resourceTypes:
+  collection: !include raml-util/rtypes/collection.raml
+  collection-item: !include raml-util/rtypes/item-collection.raml
+
+/scheduled-notice-storage:
```
marcjohnson-kint: I think both the post and put endpoints need to declare the `validate` trait, as they can respond with validation errors.

For [example](https://github.com/folio-org/mod-inventory-storage/blob/b8e5b4f13594ce22fede6e1e4c1a8a92b6508834/ramls/location.raml#L43)

Maybe, in future, the standard resource type will be changed to include this trait all of the time

ramls/scheduled-notice.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "id": {
```
marcjohnson-kint: This should also have the UUID validation regular expression

ramls/scheduled-notice-storage.raml
opochepam: Thank you. I thought the standard resource type includes validate trait.

Added the trait for post and put.

ramls/scheduled-notice.json
opochepam: added the expression

ramls/scheduled-notice-storage.raml
marcjohnson-kint: Thanks

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
```diff
@@ -0,0 +1,162 @@
...
+  @Override
+  public void deleteScheduledNoticeStorageScheduledNotices(String query,
+                                                           Map<String, String> okapiHeaders,
+                                                           Handler<AsyncResult<Response>> asyncResultHandler,
+                                                           Context vertxContext) {
+    
+      
+      PostgresClient pgClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+
+      clqToSqlDeleteQuery(query, okapiHeaders.get(RestVerticle.OKAPI_HEADER_TENANT))
```
marcjohnson-kint:
```suggestion
      cqlToSqlDeleteQuery(query, okapiHeaders.get(RestVerticle.OKAPI_HEADER_TENANT))
```

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
```diff
@@ -0,0 +1,162 @@
...
+                                                                           Handler<AsyncResult<Response>> asyncResultHandler,
+                                                                           Context vertxContext) {
+
+    PgUtil.put(SCHEDULED_NOTICE_TABLE, entity, scheduledNoticeId, okapiHeaders, vertxContext,
+      PutScheduledNoticeStorageScheduledNoticesByScheduledNoticeIdResponse.class, asyncResultHandler);
+
+  }
+
+  private Future<Void> executeSql(PostgresClient pgClient, String sql) {
+
```
marcjohnson-kint: Why is there a blank line here?

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
```diff
@@ -0,0 +1,162 @@
...
+    Future<UpdateResult> future = Future.future();
+    pgClient.execute(sql, future.completer());
+    return future.map(ur -> null);
+  }
+
+  private Future<String> clqToSqlDeleteQuery(String cql, String tenant) {
+
```
marcjohnson-kint: Similarly, why is there a blank line here?

src/test/java/org/folio/rest/api/ScheduledNoticesAPITest.java
```diff
@@ -0,0 +1,308 @@
...
+
+    assertThat(scheduledNotices.getScheduledNotices().size(), is(2));
+    assertThat(scheduledNotices.getTotalRecords(), is(2));
+  }
+
+  @Test
+  public void canGetScheduledNoticesCollectionByQuery() throws MalformedURLException, InterruptedException,
+    ExecutionException, TimeoutException {
+
+    String templateId = "b84b721e-76ca-49ef-b486-f3debd92371d";
```
marcjohnson-kint: I prefer the use of dynamic UUIDs, however I'm aware many folks prefer the hard coded approach

src/test/java/org/folio/rest/api/ScheduledNoticesAPITest.java
```diff
@@ -0,0 +1,308 @@
...
+    client.delete(scheduledNoticesStorageUrl("/scheduled-notices?" + query), TENANT_ID, ResponseHandler.empty(deleteCompleted));
+    deleteCompleted.get(5, SECONDS);
+
+    CompletableFuture<JsonResponse> getCompleted = new CompletableFuture<>();
+    client.get(scheduledNoticesStorageUrl("/scheduled-notices?" + query), TENANT_ID, ResponseHandler.json(getCompleted));
+    JsonResponse response = getCompleted.get(5, SECONDS);
+
+    ScheduledNotices scheduledNotices = response.getJson().mapTo(ScheduledNotices.class);
+
+    assertThat(scheduledNotices.getScheduledNotices().size(), is(0));
```
marcjohnson-kint: Hmm, this test appears to demonstrate deleting notices by `templateId`.

Two out of the three notices use that template, and the third record should still exist.

Why does it not check that the third notice still exists?

This test would pass even if all of the notices were deleted, which I imagine is not the desired behaviour.

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
opochepam: For me it seems easier to read, when method's signature and body are split by a blank line.

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
opochepam: fixed

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
opochepam: I see different examples in mod-circulation-storage. Some methods have a blank line, but some don't.

It would be good to keep coding style consistency for the module. Probably it is better to remove a blank line. What do you think?

src/test/java/org/folio/rest/api/ScheduledNoticesAPITest.java
opochepam: I see that in most of the test in mod-circulation-storage random values are used. So I made a changes to use random values in order to keep consistency with other tests.

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
marcjohnson-kint: I personally think that decision is highly contextual.

If a method has lots of parameters, I find the whitespace often helps to distinguish the definition from the implementation.

If a method is small and simple, I find the whitespace distracting.

I think agreeing on a consistent set of guidance would be useful. That is likely a broader topic of discussion (and likely relates to method parameter indentation, which is also important and inconsistent).

src/test/java/org/folio/rest/api/ScheduledNoticesAPITest.java
marcjohnson-kint: Thank you

src/test/java/org/folio/rest/api/ScheduledNoticesAPITest.java
opochepam: Added a check to verify that third notice still exists.

PR #170, opened 2019-05-29
[CIRCSTORE-133] Request Expiration Job Takes Too Long for Short Term Expiry Periods

kostyantyn-kh:
reduce run period to "every 2 minutes" and process up to 150 expired requests

### DIFF
--- a/src/main/java/org/folio/support/ExpirationTool.java
+++ b/src/main/java/org/folio/support/ExpirationTool.java
@@ -88,7 +88,8 @@ public class ExpirationTool {
 
     String where = format("WHERE " +
         "(jsonb->>'status' = '%1$s' AND jsonb->>'requestExpirationDate' < '%3$s') OR " +
-        "(jsonb->>'status' = '%2$s' AND jsonb->>'holdShelfExpirationDate' < '%3$s')",
+        "(jsonb->>'status' = '%2$s' AND jsonb->>'holdShelfExpirationDate' < '%3$s') " +
+        "LIMIT 150",
 
       OPEN_NOT_YET_FILLED.value(),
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -88,7 +88,8 @@ private ExpirationTool() {
 
     String where = format("WHERE " +
         "(jsonb->>'status' = '%1$s' AND jsonb->>'requestExpirationDate' < '%3$s') OR " +
-        "(jsonb->>'status' = '%2$s' AND jsonb->>'holdShelfExpirationDate' < '%3$s')",
+        "(jsonb->>'status' = '%2$s' AND jsonb->>'holdShelfExpirationDate' < '%3$s') " +
+        "LIMIT 150",
```
marcjohnson-kint: Why has this limit been introduced?

kostyantyn-kh: To process limited amount of requests in one run, prevents potential overlaps between runs

marcjohnson-kint: How was 150 decided upon as an appropriate number?

kostyantyn-kh: 150 gives us 1050 processed requests every quarter of an hour.
Tested locally with 10 requests queues * 150, average process time <10s.
Still request queue size is unpredictable, there are only average values available.
So i've included potential processing delays, possible long queues etc.

marcjohnson-kint: Fair enough

PR #172, opened 2019-06-11
CIRCSTORE-134: Update RMB to version 25.0.0

julianladisch:
* Replace duplicated code by PgUtil.get, PgUtil.getById, PgUtil.put, PgUtil.post, PgUtil.deleteById
* Replace duplicated post/put/upsert code by MyPgUtil.putUpsert204, note that this violates the HTTP spec by returning 204 on insert, it should be 201!
* Replace deprecated PostgresClient::mutate by PostgresClient::execute where possible
* Replace SQL gen_random_uuid() by deterministic calculation (see RMB-347)
* Using PgUtil.* also fixes validations errors that returned 500 but now return 400.
* Drop updateRequestWithoutUserHeaderFails() test, this no longer fails and need to be checked by Okapi.
* HttpResponseStatusCodeMatchers: Fix isUnprocessableEntity(), add isInternalServerError().
* Fix TextDateTimeMatcher to match if both timestamps are equal.

Not changed:
* PutStaffSlipsStorageStaffSlipsByStaffSlipId still returns 201 status on insert, this is missing from the .raml file!
* LoansApiHistoryTest fails (disabled using `@Ignore`), it requires an RMB audit table bugfix: https://issues.folio.org/browse/RMB-402

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>domain-models-runtime</artifactId>
-      <version>24.0.0</version>
+      <version>25.0.0</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>domain-models-runtime</artifactId>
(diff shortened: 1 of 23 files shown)


marcjohnson-kint:
Closing this as there is work underway to upgrade to RAML Module Builder 26 instead

PR #174, opened 2019-07-04
CIRCSTORE-134: Update to RMB 26.1.0

julianladisch:
Breaking change:
* New loan history (audit) structure

Other changes:
* Replace duplicated code by PgUtil.get, PgUtil.getById, PgUtil.put, PgUtil.post, PgUtil.deleteById
* Replace duplicated post/put/upsert code by MyPgUtil.putUpsert204, note that this violates the HTTP spec by returning 204 on insert, it should be 201!
* Replace deprecated PostgresClient::mutate by PostgresClient::execute where possible
* Replace SQL gen_random_uuid() by deterministic calculation (see RMB-347)
* Using PgUtil.* also fixes validations errors that returned 500 but now return 400.
* Drop updateRequestWithoutUserHeaderFails() test, this no longer fails and need to be checked by Okapi.
* HttpResponseStatusCodeMatchers: Fix isUnprocessableEntity(), add isInternalServerError().
* Fix TextDateTimeMatcher to match if both timestamps are equal.

Not changed:
* PutStaffSlipsStorageStaffSlipsByStaffSlipId still returns 201 status on insert, this is missing from the .raml file!


ramls/request.json
```diff
@@ -70,17 +70,7 @@
...
-          "description": "title of the item",
-          "type": "string"
-        },
-        "barcode": {
-          "description": "barcode of the item",
-          "type": "string"
-        }
-      },
-      "additionalProperties": false
+      "$ref": "request-itemdata.json"
```
marcjohnson-kint: Why has this been extracting into a separate fragment?

ramls/loans-history-items.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Collection of loan history items",
+  "type": "object",
+  "properties": {
+    "loans-history": {
```
marcjohnson-kint: I believe we usually use camel case for JSON property names. Is there a reason why this is hyphenated instead?

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -920,33 +920,6 @@ public void updatedRequestHasUpdatedMetadata()
       metadata.getString("updatedDate"), is(not(createdMetadata.getString("updatedDate"))));
   }
 
-  @Test
```
marcjohnson-kint: Has this behaviour changed?

ramls/request.json
julianladisch: There was a name clash with the audit table 'item' property field: https://github.com/folio-org/raml-module-builder/pull/455
This has been resolved, now the audit table property field name needs to be configured in schema.json "auditingFieldName".
I moved request-itemdata.json back into request.json

ramls/loans-history-items.json
julianladisch: No, changed.

ramls/request.json
marcjohnson-kint: @julianladisch Thanks for sharing the background to this

src/test/java/org/folio/rest/api/RequestsApiTest.java
julianladisch: Yes.
The user id is no longer a required field in the metadata because trying to login should work without a user id.
https://github.com/folio-org/raml/pull/120/files
https://github.com/folio-org/raml-module-builder/pull/414/files

Therefore we no longer get a 500 but a 204 response code for a put where user id is null.

If we don't trust Okapi to only invoke modules when the headers are correct we may add some header validation to RMB RestVerticle, but this is a separate issue: https://issues.folio.org/browse/RMB-423
Okapi is in charge of checking whether calling a module without a valid user id 
https://github.com/folio-org/okapi/pull/294/files

marcjohnson-kint: Thanks for explaining the thought process.

I agree that trusting clients and treating input defensively are separate and bigger topics, for another day.

PR #182, opened 2019-07-23
CIRCSTORE-138: request status filter

wwelling:
refactor request status enum replacing hyphen with colon

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,7 +32,7 @@
 * Introduces request policies (CIRCSTORE-97)
 * Introduces initial patron notice policies (CIRCSTORE-89, CIRCSTORE-90, CIRCSTORE-91, CIRCSTORE-92, CIRCSTORE-93, CIRCSTORE-100, CIRCSTORE-111, CIRCSTORE-112) 
 * Renames loan rules storage to circulation rules storage (CIRCSTORE-102)
-* Introduces `Closed: Unfilled`, `Closed: Pickup expired` and `Open: In transit` request states ( CIRCSTORE-101, CIRCSTORE-109, CIRCSTORE-110)
+* Introduces `Closed - Unfilled`, `Closed - Pickup expired` and `Open - In transit` request states ( CIRCSTORE-101, CIRCSTORE-109, CIRCSTORE-110)
 * Adds `opening time offset period` to loan policy (CIRCSTORE-64)
 * Adds `tags` to requests (CIRCSTORE-99)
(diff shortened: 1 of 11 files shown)


marcjohnson-kint:
I really don’t think we want to change the definition of request states to address this issue. 

It is a breaking compatibility change and directly affects the presentation and domain language.

wwelling:
I agree @marcjohnson-kint. I only present this as a working solution. However, if there is no practical way to afford hyphens in comprehensive string matching or there is only an egregious hack, then it is our prerogative to inform product owners of the limitations of the technology as for them to decide on other notation. Also, a coordinated search and replace is all that is required. That is if the product owners and stakeholder accept the refactor.

wwelling:
I do believe the commits show some strategies that don’t work and may lead to require looking in raml-module-builder for a solution.

marcjohnson-kint:
> I agree @marcjohnson-kint. I only present this as a working solution. 

Sure, and I'm thankful for the investigation. Apologies if my comment last night was rather direct, it was late and I didn't want anyone to enthusiastically merge this change given it had been approved.

> However, if there is no practical way to afford hyphens in comprehensive string matching or there is only an egregious hack, then it is our prerogative to inform product owners of the limitations of the technology as for them to decide on other notation. 

The `==` relation in CQL performs behaves as expected. This is what mod-circulation uses for status filtering. Other clients could use this (it could be argued that the semantics are a better fit). 
There was hope that there could be a quick backend fix to give the UI more time to transition to this, as it is unlikely to be trivial.

> Also, a coordinated search and replace is all that is required. That is if the product owners and stakeholder accept a refactor. 

Sure, I'd prefer the kind of upheaval of a breaking change for this reason if possible. That would not represent the quick-fix which was the intention behind this work. If necessary, we can have that conversation with folks.

> Additionally, a simple string enum doesn’t necessary have to be the gloss presentation, but rather a means to communicate the meaning. It should be up to the UI to convert to presentation and not require it be the value of the enum.

Of course, the string enum approach to states is far from ideal, localization being only one of the challenges. 

As far as I'm aware, FOLIO has not determined a way to model states that need to be referred to in processes, as reference records.

> I do believe the commits show some strategies that don’t work and may lead to require looking in raml-module-builder for a solution.

That they do. I'm glad you shared them to demonstrate the approaches tried.

wwelling:
@marcjohnson-kint I believe the label `do not merge` can be reevaluated.

PR #185, opened 2019-07-29
Anonymize list of loans

mrdpopov:
## Purpose
Currently there is only one method that anonymizes all closed loans by userId. To fulfill the requirements of CIRC-368, we need to selectively anonymize only those loans that do not have a fee/fine assigned to them. This endpoint will also be useful in other parts of the system where it is expected to anonymize a list of loans, not all loans for a particular user. For example, it's planned to be re-used in mod-circulation to fullfil the requirements of the functionality of automatic anonymization: UXPROD-1085.

https://issues.folio.org/browse/CIRCSTORE-137
Followup ticket https://issues.folio.org/browse/CIRC-368

## Approach
The existing functionality of anonymizing all loans for a particular user has been taken as a baseline with additional tweaks: 

Method: POST
Path: /anonymize-storage-loans
Returns: 
- http status 200 with json body schema:
```
{
  "anonimizedLoans": [list of ids],
  "notAnonimizedLoans": [reason, [list of ids]]
}
```
- no restriction for the input params size. It is the clients's job to batch their data if needed, currently  there's no even ballkpark estimate of the number of loans to be anonymized. 
- for any other error: http status 500 and the message of error.

The method will only anonymize loans with `closed status` that have a userId assosiation as well as update the records in the _audit_loan_ table.


ramls/loan-storage.raml
```diff
@@ -28,6 +29,27 @@ resourceTypes:
...
+      body:
+        application/json:
+          type: string[]
+      responses:
+        200:
+          description: "Number of closed loans for this user that have been anonymized"
+          body:
+            application/json:
+              type: anonymize-loans-response
+        422:
```
opochepam: 422 response is redundat here as it is already included in `validate` (see validation.raml)

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,63 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+          Json.encode(invalidUuids),
+          "Loan IDs should be a valid UUID");
+        responseHandler.handle(succeededFuture(
+          PostLoanStorageLoansAnonymizeByUserIdResponse
+            .respond422WithApplicationJson(errors)));
+        return;
+      }
+
+      final String tenantId = TenantTool.tenantId(okapiHeaders);
+      final PostgresClient postgresClient = PostgresClient.getInstance(
```
opochepam: can be simplified with `final PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, okapiHeaders);`

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,63 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+        vertxContext.owner(), tenantId);
+
+      final String combinedAnonymizationSql = createAnonymizationSQL(loanIds,
+        tenantId);
+      log.info(String.format("Anonymization SQL: %s", combinedAnonymizationSql));
+      update(postgresClient, combinedAnonymizationSql)
+        .map(updateResult -> {
+          response.setUpdatedCount(updateResult.getUpdated());
```
opochepam: `response.withUpdatedCount` can be used here to shorten the lambda expression to one line

mrdpopov:
Tests are being worked on.

descriptors/ModuleDescriptor-template.json
```diff
@@ -370,6 +375,11 @@
       "displayName": "Circulation storage - anonymize loans for a user",
       "description": "Anonymize closed loans for a single user"
     },
+    {
+      "permissionName": "circulation-storage.anonymizeloans.post",
```
opochepam: please add the permission to `subPermissions` of `circulation-storage.all`

ramls/loan-storage.raml
mrdpopov: Done.

src/main/java/org/folio/rest/impl/LoansAPI.java
mrdpopov: Done.

src/main/java/org/folio/rest/impl/LoansAPI.java
mrdpopov: Done.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,57 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+      .map(updateResult -> PostLoanStorageAnonymizeLoansResponse.
+        respond200WithApplicationJson(response.withUpdatedCount(
+          updateResult.getUpdated())))
+      .map(Response.class::cast)
+      .otherwise(this::mapExceptionToResponse)
+      .setHandler(responseHandler);
+
+  }
+
+  private Response mapExceptionToResponse(Throwable t) {
```
opochepam: Seems like this method is redundant. Just use return statement in otherwise section.

mrdpopov: Done.

mrdpopov:
The request schema has been changed form string[] to {loandIds: string[]}. Please advice if this is a better approach.

opochepam:
> The request schema has been changed form string[] to {"loanIds": string[]}. Please advice if this is a better approach.

mrdpopov:
It's in mod-circulation-storage because the existing approach was modifying the database directly and this PR tried to follow that way. In addition to removing userId from loans, it is also removing  the record from the _audit_loan_ table, which, I believe, can be done in this module only. Please correct me, if I'm wrong.

If were were to update each loan individually with PUT requests, I believe, the following are qualification against this approach:
1. Reduced performance because HTTP requests will bounce between services and okapi, which leads to:
2. Expected increase in the network traffic. 
The above points will be especially noticeable for large arrays of loans.
3. Anonimizing a single loan is almost the same as anonimizing a list of loans: a single SQL update.
3. "Obsolete" record will remain in the _audit_loan_ table.

In general, I think, anonimizing loans does not look like a `batch` because it's done in a single SQL update, even for larger inputs.

Anonymization is indeed moving to mod-circulation. This endpoint is planned to be used in CIRC-368

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,54 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+                                            Context vertxContext) {
+
+    AnonymizeLoansResponse response = new AnonymizeLoansResponse();
+
+    List<String> loanIds = request.getLoanIds();
+    final Set<String> invalidUuids = loanIds.stream()
+      .filter(id -> !UUIDValidation.isValidUUID(id))
+      .collect(Collectors.toSet());
+
+    if (!invalidUuids.isEmpty()) {
```
kostyantyn-kh: Is it intended to cancel loans anonymization for all loans in case some of loan ids are incorrect?

mrdpopov: Yes

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,54 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+        PostLoanStorageLoansAnonymizeByUserIdResponse
+          .respond422WithApplicationJson(errors)));
+      return;
+    }
+
+    final String tenantId = TenantTool.tenantId(okapiHeaders);
+    final PostgresClient postgresClient = PgUtil.postgresClient(vertxContext,
+      okapiHeaders);
+
+    final String combinedAnonymizationSql = createAnonymizationSQL(loanIds,
```
kostyantyn-kh: Could you add check for empty loandIds?

src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -68,6 +69,43 @@ public void checkIdsAfterEach() {
...
 
+  @Test
+  public void canAnonymizeLoans() throws InterruptedException, ExecutionException, TimeoutException, MalformedURLException {
+
+    UUID itemId = UUID.randomUUID();
+    UUID userId = UUID.randomUUID();
+
+    JsonObject loan1 = loansClient.create(new LoanRequestBuilder()
+      .withId(UUID.randomUUID())
+      .withItemId(itemId)
```
kostyantyn-kh: Is it possible to replace _itemId_ and _userId_ with UUID.randomUUID()?
Or do you plan to use these variables somewhere in the code?

ramls/anonymize-loans-request.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Request schema for anonymize loans",
+  "type": "object",
+  "properties": {
+    "loanIds": {
```
kostyantyn-kh: Does it make sense to restrict number of loanId to be processed?

mrdpopov: I'm not sure, the list is not expected to be very large. It's going to be loans created in one session.

src/test/java/org/folio/rest/api/LoansApiTest.java
mrdpopov: Done

ramls/anonymize-loans-request.json
mrdpopov: https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause
It's stated there that the number of params in `where ... in` clause can be quite large to satisfy our needs.

marcjohnson-kint:
@mrdpopov Thank you for the context.

> It's in mod-circulation-storage because the existing approach was modifying the database directly and this PR tried to follow that way. 

That approach was deeply tactical as an expedient way to do the first manual anonymization, it is not advised as a direction or approach. 

The existing endpoint should be deprecated once it is replaced by the new mechanisms in mod-circulation.

For the new endpoint, if possible please include this in a separate interface specific to anonymization. This allows a module to choose whether to implement anonymization as well as regular loans storage.

> In addition to removing userId from loans, it is also removing the record from the _audit_loan_ table, which, I believe, can be done in this module only. Please correct me, if I'm wrong.

Can you help me understand where this requirement came from? I did a quick scan of the issues and didn't notice this requirement.

> If we were updating each loan individually with PUT requests, I believe, the following would be qualifications against this approach:
> 
> 1. Expected increase in the network traffic because HTTP requests will bounce between services and okapi, which leads to:
> 2. **Reduced performance**
>    The above points will be especially noticeable for large arrays of loans.
> 3. Anonimizing a single loan is almost the same as anonimizing a list of loans: a single SQL update.
> 4. "Obsolete" record will remain in the _audit_loan_ table.
> 
> In general, I think, anonimizing loans does not look like a `batch` because it's done in a single SQL update, even for larger inputs.
> 

I can understand why you are concerned about performance. I think tailoring the public API based upon performance concerns tends to introduce trade offs, and I like to ask questions to try to understand how those trade offs were made.

mrdpopov:
Can you help me understand where this requirement came from? I did a quick scan of the issues and didn't notice this requirement.

That was the way the old endpoint was implemented.

marcjohnson-kint:
>> Can you help me understand where this requirement came from? I did a quick scan of the issues and didn't notice this requirement.
> 
> That was the way the old endpoint was implemented.

Good answer. It is probably necessary due to the flaws in that implementation :-/

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "loan-storage",
-      "version": "6.1",
+      "version": "6.2",
```
marcjohnson-kint: If this stays in the loan-storage interface, please also update the RAML file

descriptors/ModuleDescriptor-template.json
```diff
@@ -38,6 +38,11 @@
           "methods": ["POST"],
           "pathPattern": "/loan-storage/loans/anonymize/{userId}",
           "permissionsRequired": ["circulation-storage.loans.collection.anonymize.user.post"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/loan-storage/anonymize-loans",
+          "permissionsRequired": ["circulation-storage.anonymizeloans.post"]
```
marcjohnson-kint: Personally, I'd name this as similar to the endpoint as possible, using hyphens between words.

For example,
```suggestion
          "permissionsRequired": ["circulation-storage.anonymize-loans.post"]
```

ramls/anonymize-loans-request.json
marcjohnson-kint: What happens if that is exceeded, will the client receive a server error?

ramls/anonymize-loans-response.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Response schema for anonymize loans request",
+  "type": "object",
+  "properties": {
+    "updatedCount": {
```
marcjohnson-kint: What will this property be used for?

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -1,21 +1,18 @@
...
-import java.util.ArrayList;
-import java.util.Map;
-import java.util.Objects;
-import java.util.StringJoiner;
-import java.util.function.Function;
-
-import javax.validation.constraints.NotNull;
-import javax.ws.rs.core.Response;
-
+import io.vertx.core.AsyncResult;
```
marcjohnson-kint: We typically use the eclipse style of import ordering - see https://issues.folio.org/browse/FOLIO-1667

Please update this to respect that ordering to reduce the amount of unnecessary flux between changes.

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -180,6 +186,54 @@ public void postLoanStorageLoansAnonymizeByUserId(
...
+  public void postLoanStorageAnonymizeLoans(AnonymizeLoansRequest request,
+                                            Map<String, String> okapiHeaders,
+                                            Handler<AsyncResult<Response>> responseHandler,
+                                            Context vertxContext) {
+
+    AnonymizeLoansResponse response = new AnonymizeLoansResponse();
+
+    List<String> loanIds = request.getLoanIds();
+    final Set<String> invalidUuids = loanIds.stream()
+      .filter(id -> !UUIDValidation.isValidUUID(id))
```
marcjohnson-kint: Why does it matter if the input are UUIDs or not?

mrdpopov: For optimization purposes: do not execute SQL when we know in advance uuids are invalid. They are used in the where...in clause in SQL query and  If we provide something that is not UUID, the database will throw error.

ramls/anonymize-loans-response.json
mrdpopov: It's under rework.

src/test/java/org/folio/rest/support/ApiTests.java
```diff
@@ -97,4 +98,11 @@ protected void checkNotFound(URL url)
 
     assertThat(statusCode, is(HttpURLConnection.HTTP_NOT_FOUND));
   }
+
+  protected void doesNotHaveProperty(String property, JsonObject resource, String type) {
```
marcjohnson-kint: This likely belongs in a separate place to the ApiTests base class, maybe as a custom matcher?

ramls/anonymize-loans-request.json
mrdpopov: A preliminary number of loans to be anonymized in one session will be about 1000 - a number from the PO.

src/test/java/org/folio/rest/api/LoansApiTest.java
mrdpopov: This has been reworked.

ramls/anonymize-loans-request.json
mrdpopov: If the number is greater than 1000,        it will respond with 422  
```
{
    "errors": [
        {
            "message": "size must be between 0 and 1000",
            "type": "1",
            "code": "-1",
            "parameters": [
                {
                    "key": "loanIds",
                    "value": "[list of 1000 ids]"
                }
            ]
        }
    ]
}
```

descriptors/ModuleDescriptor-template.json
mrdpopov: This has been reworked.

ramls/anonymize-loans-response.json
mrdpopov: This has been reworked.

src/main/java/org/folio/rest/impl/LoansAPI.java
mrdpopov: This has been reworked.

src/test/java/org/folio/rest/support/ApiTests.java
mrdpopov: This has been reworked.

src/main/java/org/folio/rest/impl/LoansAPI.java
mrdpopov: Done

ramls/anonymize-loans-request.json
mrdpopov: According to the new details from the PO, the number of items per sesison can be as high as 2,636. @marcjohnson-kint @kostyantyn-kh Do you think it makes sence to restrict number of loanId to be processed to 3000? Alternatively, we can remove the restriction as the number can be quite high.

marcjohnson-kint: What do you mean by items per session?

What is the limit in the underlying implementation?

Could the client, mod-circulation, submit batches of loans to anonymize?

mrdpopov: The limit of the implementation is base upon how many Ids we can put into the `where...in` clause and I think that number can be quite large.  I also  think it's best to remove the size restriction here and the clients will submit batches of IDs if needed.

descriptors/ModuleDescriptor-template.json
```diff
@@ -41,6 +41,17 @@
         }
       ]
     },
+    {
+      "id": "anonymize-loans",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/anonymize-loans",
```
marcjohnson-kint: I thought we agreed that the path would be storage specific, in order to reserve a top-level path like `/anonymize-loans` for a business logic endpoint, like manual anonymization.

Can you help me understand why this path was chosen? Was it to do with the RAML Module Builder limitations of overlapping paths?

ramls/anonymize-loans-response.json
```diff
@@ -0,0 +1,22 @@
...
+      "type": "array",
+      "items":{
+        "type": "string"
+      }
+    },
+    "notAnonimizedLoans": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "$ref": "anonymize-loans-failure.json"
```
marcjohnson-kint: Why has this been extracted as a separate schema?

src/main/java/org/folio/rest/impl/LoansAPI.java
mrdpopov: This has been reworked.

ramls/anonymize-loans-request.json
marcjohnson-kint: Does the size limit defined actually change any behaviour, e.g. if a client submits 1001 loan IDs with the current implementation, what happens?

mrdpopov: It will be processed successfully. The only limit is the database's ability to handle many items in where...in clause. More info: https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause
One of the concerns for adding this restriction was related to the table row locking. We may experience issues if the table is locked for too long because of update with many items in the where...in clause. 
I still suggest we  remove the restriction here and, if needed, the clients will batch their data.

ramls/anonymize-loans-response.json
mrdpopov: I guess that was the common practice in Folio. Shout it be inlined?

descriptors/ModuleDescriptor-template.json
mrdpopov: It has to do with the RMB issue. It looks like I cant add a POST /loan-storage/loans/anonymize endpoint if there is already POST /loan-storage/loans/anonymize{userId} defined

Which endpoint would you suggest to use?

marcjohnson-kint: Please can you raise a bug report for this in the RAML Module Builder project, using these example endpoints

I suggest using any path that includes storage and doesn't cause this conflict, for example, `anonymize-storage-loans`

This can easily be changed as the interface is marked experimental.

ramls/anonymize-loans-response.json
mrdpopov: This has been inlined.

descriptors/ModuleDescriptor-template.json
mrdpopov: The interface has been renamed to /anonymize-storage-loans

ramls/anonymize-storage-loans-response.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Response schema for anonymize loans request",
+  "type": "object",
+  "properties": {
+    "anonimizedLoans": {
```
mtraneis: typo? Should it be "anonymizedLoans"?

ramls/anonymize-storage-loans-response.json
```diff
@@ -0,0 +1,36 @@
...
+      "description": "Successfully anonymized loan ids",
+      "type": "array",
+      "items":{
+        "type": "string"
+      }
+    },
+    "notAnonimizedLoans": {
```
mtraneis: typo? Should it be "notAnonymizedLoans"?

mrdpopov: Thank you for noticing. Fixed.

ramls/anonymize-storage-loans-response.json
mrdpopov: Thank you for noticing. Fixed.

PR #186, opened 2019-08-02
[CIRCSTORE-141] Extend patron notice policy and scheduled notice schemas

opochepam:
## Purpose
Resolves: CIRCSTORE-141
1. Add `Request Expiriation` event to patron notice policy by the following path
patron-notice-policy.json -> requestNotices -> sendOptions -> sendWhen

2. Add `triggeringEvent` property to scheduled-notice json, values: Hold Expiration, Request Expiration
## Links
https://issues.folio.org/browse/CIRCSTORE-141
https://issues.folio.org/browse/CIRC-387

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "2.0",
+      "version": "0.9",
       "handlers": [
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.9",
+      "version": "2.0",
       "handlers": [
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.8",
+      "version": "0.9",
       "handlers": [
         {
           "methods": ["POST"],
@@ -285,7 +285,7 @@
     },
     {
       "id": "scheduled-notice-storage",
-      "version": "0.1",
+      "version": "0.2",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/patron-notice-policy.json
+++ b/ramls/patron-notice-policy.json
@@ -344,6 +344,7 @@
                   "Recall request",
                   "Recall loanee",
                   "Hold request",
+                  "Request Expiration",
                   "Paging request",
                   "Available",
                   "Hold Expiration",
--- a/ramls/patron-notice-policy.json
+++ b/ramls/patron-notice-policy.json
@@ -344,7 +344,7 @@
                   "Recall request",
                   "Recall loanee",
                   "Hold request",
-                  "Request Expiration",
+                  "Request expiration",
                   "Paging request",
                   "Available",
                   "Hold Expiration",
--- a/ramls/scheduled-notice.json
+++ b/ramls/scheduled-notice.json
@@ -22,6 +22,11 @@
       "format": "date-time",
       "description": "Next run time"
     },
+    "triggeringEvent": {
+      "type": "string",
+      "description": "Scheduled notice triggering event",
+      "enum": ["Hold Expiration", "Request Expiration"]
+    },
     "noticeConfig": {
       "type": "object",
       "properties": {
--- a/ramls/scheduled-notice.json
+++ b/ramls/scheduled-notice.json
@@ -25,7 +25,7 @@
     "triggeringEvent": {
       "type": "string",
       "description": "Scheduled notice triggering event",
-      "enum": ["Hold Expiration", "Request Expiration"]
+      "enum": ["Hold Expiration", "Request expiration"]
     },
     "noticeConfig": {
       "type": "object",
--- a/ramls/scheduled-notice.json
+++ b/ramls/scheduled-notice.json
@@ -25,7 +25,11 @@
     "triggeringEvent": {
       "type": "string",
       "description": "Scheduled notice triggering event",
-      "enum": ["Hold expiration", "Request expiration"]
+      "enum": [
+        "Hold expiration",
+        "Request expiration",
+        "Due date"
+      ]
     },
     "noticeConfig": {
       "type": "object",
--- a/ramls/scheduled-notice.json
+++ b/ramls/scheduled-notice.json
@@ -25,7 +25,7 @@
     "triggeringEvent": {
       "type": "string",
       "description": "Scheduled notice triggering event",
-      "enum": ["Hold Expiration", "Request expiration"]
+      "enum": ["Hold expiration", "Request expiration"]
     },
     "noticeConfig": {
       "type": "object",
--- a/ramls/patron-notice-policy.raml
+++ b/ramls/patron-notice-policy.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Patron Notice Policies
-version: v0.8
+version: v0.9
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost:9130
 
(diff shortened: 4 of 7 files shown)


ramls/scheduled-notice.json
```diff
@@ -22,6 +22,11 @@
       "format": "date-time",
       "description": "Next run time"
     },
+    "triggeringEvent": {
```
marcjohnson-kint: How i this intended to be used?

Why is this discriminator needed for expiration notices and not for other kinds of notices?

opochepam: It is intented for scheduled notices selection during request time based notices processing implemented in scope of this story https://issues.folio.org/browse/CIRC-387.

We also have due date event for timebased notices, currently it is not presented in the schema and is not used by due date scheduled notices processing job. It means that during due date scheduled notices processing all scheduled notices(request and due date) are retrieved by the job and the filtered by `loanId` property presence. The due date processing job is going to be reworked in scope of this story https://issues.folio.org/browse/CIRC-408 to use new due date triggering event. Due date triggering event will be added to the schema in scope of this story https://issues.folio.org/browse/CIRCSTORE-140.

mtraneis: What about reminder notices? Is it a "due date" event or something else?

opochepam: What do you mean by reminder notices? Currently the only one triggering event for loan related time-based notices is `due date` (it may have different notice configurations, e.g. upon due date, before due date, after due date.

ramls/patron-notice-policy.json
```diff
@@ -344,6 +344,7 @@
                   "Recall request",
                   "Recall loanee",
                   "Hold request",
+                  "Request Expiration",
```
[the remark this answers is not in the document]
opochepam: Yes, it's case sensitive. It has been decided to capitalize the first letter of the second word, because it had been done similarly for `Hold Expiration` option in patron-notice-policy json schema (https://github.com/folio-org/mod-circulation-storage/blob/master/ramls/patron-notice-policy.json#L349).

ramls/scheduled-notice.json
mtraneis: As described in scenario 1's of various request types in [CIRC-387](https://issues.folio.org/browse/CIRC-387), there's a reminder notice to be sent. How is this notice type is mapped to the triggering events?

descriptors/ModuleDescriptor-template.json
```diff
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.8",
+      "version": "0.9",
```
mreno-EBSCO: Should an addition to an enumeration be considered a breaking change?
If we consider an `enum` to contain the entire universe of values, which is why it can be a constant in most languages, then adding `Request Expiration` to the `enum` would change the universe. Clients may not expect this new value, and, if treated like a Java enum, will not be able to map this value to an existing v0.8 value, likely causing an error. An `enum` isn't a `list` or `array` and I don't think we should treat it as such when considering backwards compatibility.

ramls/scheduled-notice.json
opochepam: From what I understand in first scenario reminder notice is not a specific triggering event. For example "it's nearing hold expiration" means that triggering event is `Hold Expiration` with before notice config.

ramls/patron-notice-policy.json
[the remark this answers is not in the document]
mreno-EBSCO: It seems to me, `Hold Expiration` is the outlier here and that it should be corrected and not to further pollute the enumeration with mixed case second words.

opochepam:
> Can some unit tests be created that ensure these additions are correctly generated in the interface/implementation and to help prevent future regressions?
> 
> It distresses me that we are modifying schemas without providing unit tests.

Do you mean the tests checking that generated POJO's matches json schema?

ramls/scheduled-notice.json
mtraneis: Can you explain how "before notice config" works? I am not familiar with events and schedulings. If you can just provide a link to this configuration that would be great as well.

descriptors/ModuleDescriptor-template.json
opochepam: It makes sense. Changed interface version to `2.0`.

mreno-EBSCO:
> Do you mean the tests checking that generated POJO's matches json schema?

Not the schema, but the expectation. We are blindly letting code get written to and read from the DB based on the schema. If the schema changes again, we have no tests to ensure that the changes will break existing expectations. If someone lowercases the `E` in "Request Expiration` everything will build and deploy just fine, but will likely break at run time.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: @mreno-EBSCO that’s a good question, one that likely deserves a considered response from me later.

This interface was pre 1.0, and therefore explicitly unstable, why choose now to not only mark it stable and to skip the first major version?

opochepam: @marcjohnson-kint  apologies, skiped the first major version by mistake.
It seems like `feeFineNotices` is not going to be used as there is different for notice configuration mechanism presented in mod-feefines. Probably it's better desicion to leave the version under 1.0. What do you think?

ramls/scheduled-notice.json
opochepam: @mtraneis here is the link https://github.com/folio-org/mod-circulation-storage/blob/master/ramls/scheduled-notice.json#L25
Notice configuration is set based on notice policy when scheduled notice is created. For example when request with expiration date is created and there is notice policy configured for request expiration triggering event.

ramls/patron-notice-policy.json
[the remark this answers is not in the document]
opochepam: @mreno-EBSCO in this case corresponding changes in the ui module must be done as ui can get an error trying to interact with `patron-notice-policy-storage` interface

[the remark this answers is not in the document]
mreno-EBSCO: Sure, I realize that there would be changes required by clients of this API, but I believe striving for consistency is crucial to quality APIs. Maybe we need another JIRA to make the enumerations values consistent with respect to case... or we can just ignore this inconsistency. What do you think @marcjohnson-kint?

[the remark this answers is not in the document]
mtraneis: Are there any changes required in the ui module for Request Expiration if we changed it to lower case "e"?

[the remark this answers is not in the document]
opochepam: No, it can be safely changed. Request expiration option is not presented on the ui yet.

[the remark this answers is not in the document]
mtraneis: Can we change it to lower case "e" now that we know that upper casing the first letter of the second or subsequent words in the enum is not recommended?  We could deal with 'Hold Expiration' in another JIRA, as @mreno-EBSCO suggested.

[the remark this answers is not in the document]
marcjohnson-kint: @mreno-EBSCO My understanding is that FOLIO decided to standardise on first word capitalization when defining states / statuses / enumerated values e.g. `Checked out`.

With the exception that the first word after hyphenation is also capitalized e.g. `Open - Awaiting pickup`.

I would prefer we be consistent if possible without incurring significant coordination overhead (e.g. if we need to break a big, heavily used interface).

It is decisions like this where I'm advocating for many interfaces to spend some time as experimental before 1.0.

descriptors/ModuleDescriptor-template.json
```diff
@@ -223,7 +223,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.8",
+      "version": "2.0",
```
marcjohnson-kint: Why are we choosing now to go from an experimental interface version to a regular semantic version?

And why skip to 2.0, rather than 1.0?

marcjohnson-kint:
There don't appear to be any tests that demonstrate how the new properties are intended to be used. 

Please could either existing tests be changed or new tests added to demonstrate examples of intended real usage.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: @opochepam Yes, personally I would leave this interface experimental, especially given the feedback being generated as we apply further changes

opochepam:
@mreno-EBSCO @marc-johnson Updated existing tests to reflect the use of introduced properies. Also added `Due date` triggernig event as due date scheduled notice sending process is going to be modified and will require `Due date` event for notices selection.

PR #191, opened 2019-08-29
[CIRCSTORE-146] Add userId property to scheduled notice schema

dmytrotur:
## Purpose 
Add userId property to scheduled notice schema
Resolves: [CIRCSTORE-146](https://issues.folio.org/browse/CIRCSTORE-146)

### DIFF
--- a/ramls/scheduled-notice.json
+++ b/ramls/scheduled-notice.json
@@ -17,7 +17,7 @@
       "description": "UUID of related request for request notices",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
     },
-    "recipientId": {
+    "recipientUserId": {
       "type": "string",
       "description": "Id of the user to whom this notice should be sent to",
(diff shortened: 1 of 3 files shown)


ramls/scheduled-notice.json
```diff
@@ -17,6 +17,11 @@
       "description": "UUID of related request for request notices",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
     },
+    "userId": {
```
marcjohnson-kint: Could this be given a name that better reflects the role the user is playing.

When I read the comment, it sounds like this is the recipient user?

dmytrotur: Yes, makes sense for me. I'm thinking about "recipientId". What do you think, does it reflects the purpose of this field?

marcjohnson-kint: My personal preference would be `recipientUserId` as that describes both the role and the kind of record we expect the reference to refer to.

dmytrotur: Agree, changed to "recipientUserId"

PR #192, opened 2019-09-03
CIRCSTORE-150 - Delivery requests: add new staff slip - "Request deli…

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/CIRCSTORE-150 we want to add an additional staff slip entry.

## Approach
* added new staff slip entry with name "Request delivery" as a separate file.

### DIFF
--- a/reference-data/staff-slips-storage/staff-slips/requestDelivery.json
+++ b/reference-data/staff-slips-storage/staff-slips/requestDelivery.json
@@ -1,6 +1,6 @@
 {
   "id": "1ed55c5c-64d9-40eb-8b80-7438a262288b",
-  "name": "Request Delivery",
+  "name": "Request delivery",
   "active": true,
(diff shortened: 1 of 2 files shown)


reference-data/staff-slips-storage/staff-slips/requestDelivery.json
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "1ed55c5c-64d9-40eb-8b80-7438a262288b",
+  "name": "Request Delivery",
```
dmytrotur: Why does the second word start with an upper case letter? Is there a requirement for this?

natali-zaitseva: @dmytrotur, there is no particular request for capitalizing the second word, will fix it.

PR #194, opened 2019-09-06
CIRCSTORE-157: add itemEffectiveLocationAtCheckOut to loan

rjruizes:
Needed for https://github.com/folio-org/mod-circulation/pull/339

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -21,9 +21,10 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
-    "itemLocationId": {
+    "itemEffectiveLocationAtCheckOut": {
       "description": "The effective location, at the time of checkout, of the item loaned to the patron. This is a UUID that refers to /locations/{id}.",
-      "type": "string"
+      "type": "string",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
     },
     "status": {
       "description": "Overall status of the loan",
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -21,6 +21,10 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
+    "itemLocationId": {
+      "description": "Effective Location ID of the item lent to the patron",
+      "type": "string"
+    },
     "status": {
       "description": "Overall status of the loan",
       "type": "object",
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -22,9 +22,9 @@
       "type": "string"
     },
     "itemEffectiveLocationAtCheckOut": {
-      "description": "The effective location, at the time of checkout, of the item loaned to the patron. This is a UUID that refers to /locations/{id}.",
+      "description": "The effective location, at the time of checkout, of the item loaned to the patron.",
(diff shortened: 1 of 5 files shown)


nassibnassar:
Suggest for the description:

"The effective location, at the time of checkout, of the item loaned to the patron.  This is a UUID that refers to /locations/{id}."

(If that is the right target interface)

ramls/loan.json
```diff
@@ -21,6 +21,10 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
+    "itemLocationId": {
```
marcjohnson-kint: Under what circumstances is this the item location ID?

For example, is this the effective location at the point of check in, check out or the current one?

marcjohnson-kint: Based upon the description, this is the effective location at the point of check out. Please could it be renamed to reflect that.

```suggestion
    "itemEffectiveLocationAtCheckOut": {
```

marcjohnson-kint: Is this the only property we are intending to store for reporting purposes? 

If other parts of the item location is intended to be kept, or other properties about the item, it might be worth created a more nested structure.

mreno-EBSCO: Should we have the UUID pattern applied to this field? It is part of other IDs in this file and the description states that it is a UUID.

ramls/loan.json
```diff
@@ -21,6 +21,10 @@
...
+      "description": "The effective location, at the time of checkout, of the item loaned to the patron. This is a UUID that refers to /locations/{id}.",
+      "type": "string"
```
marcjohnson-kint: (new) UUID properties typically use a regular expression for validation.

The [current chosen regular expression](https://dev.folio.org/guides/uuids/) is `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$`

ramls/loan.json
```diff
@@ -21,6 +21,11 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
+    "itemEffectiveLocationAtCheckOut": {
```
marcjohnson-kint: Is the intention to only store the effective location, or could there be other aspects of the location also stored at check out?

rjruizes: Right now, only the effective location.

marcjohnson-kint: Ok, thanks

ramls/loan.json
```diff
@@ -21,6 +21,11 @@
...
+      "description": "The effective location, at the time of checkout, of the item loaned to the patron. This is a UUID that refers to /locations/{id}.",
+      "type": "string",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
```
marcjohnson-kint: FOLIO now [defines|https://dev.folio.org/guides/uuids/] `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$` as the standard UUID regular expression (in some projects, this has been defined as a reusable type). Please change this to use the newer standard.

PR #200, opened 2019-09-09
CIRC-436: Implementation for request-preference-storage

andrii-paias:
## Purpose
Store user request preferences in mod-circulation-storage

## Approach
Add CRUD endpoints and implement them by using PgUtil class

### DIFF
--- a/ramls/request-preference.json
+++ b/ramls/request-preference.json
@@ -1,0 +1,47 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "description": "Request preference schema",
+  "properties": {
+    "id": {
+      "description": "Unique request preference ID",
+      "type": "string"
+    },
+    "userId": {
+      "description": "UUID of user associated with this request preference",
+      "type": "string"
+    },
+    "holdShelf": {
+      "description": "Whether 'Hold shelf' option is available to the user.",
+      "type": "boolean"
+    },
+    "delivery": {
+      "description": "Whether 'Delivery' option is available to the user.",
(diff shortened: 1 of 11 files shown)


ramls/request-preference.json
```diff
@@ -0,0 +1,47 @@
...
+    "id": {
+      "description": "Unique request preference ID",
+      "type": "string"
+    },
+    "userId": {
+      "description": "UUID of user associated with this request preference",
+      "type": "string"
+    },
+    "holdShelf": {
+      "description": "Whether 'Hold shelf' option is available to the user.",
```
sduvvuri-ebsco: Ah, I think I misunderstood this. My understanding was that its whether "holdShelf" or "delivery" option was selected by the user - not whether which of these options is available to user.

ramls/request-preference.json
```diff
@@ -0,0 +1,47 @@
...
+    "metadata": {
+      "description": "Metadata about creation and changes to request preference",
+      "$ref": "raml-util/schemas/metadata.schema",
+      "readonly": true
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "userId",
+    "holdShelf",
```
sduvvuri-ebsco: holdShelf is required by default, is delivery required as well?

dmtkachenko: yes, it has to defined also to understand whether delivery = true or false

andrii-paias: holdShelf is supposed to be always be true (i.e. UI checkbox is always checked), delivery can be false, but I think it still makes sense to always have this field in request

PR #202, opened 2019-09-12
[CIRCSTORE-155] Change patron notice policy schema 

roman-barannyk:
## Purpose 
Move "Recall loanee" triggering event option from request notices to loan notices and rename it to "Item recalled". 
Resolves story: [CIRCSTORE-155](https://issues.folio.org/browse/CIRCSTORE-155)
It is needed for: [UICIRC-279](https://issues.folio.org/browse/UICIRC-279)

## Approach
- Changed patron notice policy schema due to requirements
- Given PR already introduces breaking change, removed "Overdue" option from loan notices as not used and renamed "Hold Expiration" to "Hold expiration" to follow convention
- Updated tests according to schema changes

## Notes 
This PR should be merged together with mod-circulation PR and ui-circulation PR (not ready yet)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -234,7 +234,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.10",
+      "version": "1.0",
       "handlers": [
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -234,7 +234,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "1.0",
+      "version": "0.11",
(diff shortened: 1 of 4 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -234,7 +234,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.10",
+      "version": "1.0",
```
marcjohnson-kint: Why is this changing to version 1.0? That would suggest the API is considered stable. My understanding is that it is still experimental

dmytrotur: Seems like there is no reason for this. We'll change it to `0.11`

roman-barannyk: The version has been updated to 0.11

PR #204, opened 2019-09-13
CIRCSTORE-156: add a flag for recall due date modification

mreno-EBSCO:
This is the storage counterpart for CIRC-440. This will persist the `dueDateChangedByRecall` in the loan. This flag will indicate to circulation that a recall has been placed that has modified the due date of the loan.

### DIFF
--- a/src/test/java/org/folio/rest/api/LoansApiTest.java
+++ b/src/test/java/org/folio/rest/api/LoansApiTest.java
@@ -1,5 +1,6 @@
 package org.folio.rest.api;
 
+import static java.lang.Boolean.TRUE;
 import static org.folio.rest.support.matchers.HttpResponseStatusCodeMatchers.isBadRequest;
 import static org.folio.rest.support.matchers.HttpResponseStatusCodeMatchers.isNotFound;
 import static org.folio.rest.support.matchers.LoanMatchers.isClosed;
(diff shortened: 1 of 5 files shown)


src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -123,6 +124,42 @@ public void canCreateALoan()
     //and presents the offset as +0000 (which is ISO8601 compatible, but not RFC3339)
     assertThat("due date does not match",
       loan.getString("dueDate"), is("2017-07-27T10:23:43.000+0000"));
+
+    assertThat("recall changed due date is set",
```
marcjohnson-kint: Should this be not set, as it is checking for a null value?

mreno-EBSCO: Possibly. 😄 The documentation for assertion messages is a bit vague about what the message should be. I used to use messages that matched the expectation of the assertion, like you suggest. Anatolli challenged that I should be using messages that were the opposite of the assertion since the message should indicate the reason for failure and you don't typically see the test context in the log, just these messages. When I looked into this, I saw mixed support for both styles. The code itself marks the message internally as failure reason, or something like that, which is why I ultimately decided to agree with him. Overall, I think not including a message, since most frameworks default messages are clear enough, is the advice I see the majority of the time.

Maybe I shouldn't be so precise. Perhaps "recall changed due date should be null" is clearer here?

marcjohnson-kint: This might well be me getting confused (I've asked similar questions on the circulation change)

When I read a message as part of an assert, I tend to think of it as describing the assertion (with some context of what is going on). Reading the inverse of the check in a message is confusing to me.

I personally prefer messages that reinforce what should've happened.

marcjohnson-kint: As you said, the [JUnit documentation](https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertThat(java.lang.String,%20T,%20org.hamcrest.Matcher) suggests the style where it states what is wrong.

Maybe it is me that finds it hard to read "X is set" next to the expectation that it isn't set.

marcjohnson-kint: My interpretation of XUnit Patterns suggests an [expectation describing message](http://xunitpatterns.com/Assertion%20Message.html) is preferable because the assertion framework tells the observer what has happened.

This seems like another one of those qualitative coding standard topics that are bigger than a single issue.

mreno-EBSCO: I agree 💯 percent. This is why I was taken back a bit when Anatolli suggested I had been doing it wrong this whole time. I find it confusing as hell to read tests like this, but I think they do make sense in the logs. I think I need to find a way to make the message clear both in the code and the log.

mreno-EBSCO: Though, we should be consistent. It looks like the messages in this test are about the tests themselves. I should probably do that as well or else it is even more confusing.

mtraneis: I found it confusing, too. I think a message that mentions both the expected behavior and the actual value is clearer in both the code and log, albeit a bit verbose.

mreno-EBSCO: Even with a message supplied JUnit should include the values. I think we just need to come to a consensus for what we should put in the message or if we want messages at all. I think the JUnit messages are pretty clear.

With message:
```
java.lang.AssertionError: recall changed due date is set

Expected: null
     but: was <true>
	at org.folio.rest.api.LoansApiTest.canCreateALoan(LoansApiTest.java:129)
```
Without message:
```
java.lang.AssertionError: 

Expected: null
     but: was <true>
	at org.folio.rest.api.LoansApiTest.canCreateALoan(LoansApiTest.java:129)
```

marcjohnson-kint: The primary reason I add a message is to provide some context about what is being checked.

When this is values of properties from JSON or an object, the most important thing I care about is what property is being checked. After that, I care about what circumstances were being checked, beyond the matcher itself.

This likely falls inbetween xUnit patterns assertion identifying message and expectation describing message.

Using the example from this test, I'd likely write (on a good day) something like:
`Due date should not have been changed due to recall`

PR #207, opened 2019-09-30
CIRCSTORE-159: Disallow alternateRenewalLoanPeriod and renewalsPolicy…

bohdan-suprun:
Breaking changes.

Implement validation rules to block following cases:
* Block `alternateRenewalLoanPeriod` for policies with Fixed profile;
* Block `RenewalsPolicy/Period` for policies with Fixed profile.

For more details see https://issues.folio.org/browse/CIRC-200?focusedCommentId=61428&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-61428.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 10.0.0 2019-10-03
+
+* Disallows `alternateRenewalLoanPeriod for holds` and `renewalsPolicy.period` for fixed policies (CIRCSTORE-159).
+
 ## 9.3.0 2019-09-18
 
 * Stores whether a loan `due date has changed due to a recall` request (CIRCSTORE-156)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,3 @@
-## 10.0.0 2019-10-03
-
-* Disallows `alternateRenewalLoanPeriod for holds` and `renewalsPolicy.period` for fixed policies (CIRCSTORE-159).
-
 ## 9.3.0 2019-09-18
 
 * Stores whether a loan `due date has changed due to a recall` request (CIRCSTORE-156)
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -69,7 +69,7 @@
     },
     {
       "id": "loan-policy-storage",
-      "version": "3.0",
+      "version": "2.0",
       "handlers": [
         {
           "methods": ["GET"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -69,7 +69,7 @@
     },
     {
       "id": "loan-policy-storage",
-      "version": "2.0",
+      "version": "3.0",
       "handlers": [
         {
           "methods": ["GET"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -377,7 +377,7 @@
     },
     {
       "id": "patron-action-session-storage",
-      "version": "0.1",
+      "version": "0.2",
       "handlers": [
         {
           "methods": ["GET"],
@@ -399,6 +399,10 @@
           "methods": ["DELETE"],
           "pathPattern": "/patron-action-session-storage/patron-action-sessions/{id}",
           "permissionsRequired": ["patron-action-session-storage.patron-action-sessions.item.delete"]
+        }, {
+          "methods": ["GET"],
+          "pathPattern": "/patron-action-session-storage/expired-session-patron-ids",
+          "permissionsRequired": ["patron-action-session-storage.expired-session-patron-ids.collection.get"]
         }
       ]
     }
@@ -786,7 +790,8 @@
         "patron-action-session-storage.patron-action-sessions.item.get",
         "patron-action-session-storage.patron-action-sessions.item.post",
         "patron-action-session-storage.patron-action-sessions.item.put",
-        "patron-action-session-storage.patron-action-sessions.item.delete"
+        "patron-action-session-storage.patron-action-sessions.item.delete",
+        "patron-action-session-storage.expired-session-patron-ids.collection.get"
       ]
     },
     {
--- a/src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
+++ b/src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
@@ -1,5 +1,7 @@
 package org.folio.rest.api;
 
+import static org.folio.rest.support.builders.LoanPolicyRequestBuilder.defaultRollingPolicy;
+import static org.folio.rest.support.builders.LoanPolicyRequestBuilder.emptyPolicy;
 import static org.folio.rest.support.matchers.HttpResponseStatusCodeMatchers.isBadRequest;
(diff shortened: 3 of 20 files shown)


marcjohnson-kint:
@bohdan-suprun Can you help me understand why it was chosen to implement this validation using Drools based rules rather than Java code?

descriptors/ModuleDescriptor-template.json
```diff
@@ -69,7 +69,7 @@
     },
     {
       "id": "loan-policy-storage",
-      "version": "2.0",
+      "version": "3.0",
```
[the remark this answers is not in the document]
bohdan-suprun: @marcjohnson-kint. not sure if this a breaking change, yep I only added the validation rules. However,  in case a consumer will specify `alternateRenewalLoanPeriod` or `RenewalsPolicy/period` for fixed policies that will not work longer.

bohdan-suprun:
> @bohdan-suprun Can you help me understand why it was chosen to implement this validation using Drools based rules rather than Java code?

@marcjohnson-kint, we already have set of drool rules for policies (for `alternateFixedDueDateScheduleId` and `fixedDueDateScheduleId` properties), so I decided to use the same approach here as well.

marcjohnson-kint:
> > @bohdan-suprun Can you help me understand why it was chosen to implement this validation using Drools based rules rather than Java code?
> 
> @marcjohnson-kint, we already have set of drool rules for policies (for `alternateFixedDueDateScheduleId` and `fixedDueDateScheduleId` properties), so I decided to use the same approach here as well.

Ok, that is fair enough. I personally find it harder to find and follow these validations when they are defined separately in Drools. I understand the decision to follow the precedent already set.

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
marcjohnson-kint: As I understand it, from an interface definition perspective, it is legitimate to define a policy with this policy, in the sense that nothing within the interface stops this from happening.

Given that it is not possible to express this kind of conditional validation within the interface definition (at the moment), I think this is an implementation breaking change, rather than an interface compatibility breaking change.

That means that this change should change the implementation version (in the pom.xml, from 9.x to 10.0.0 I believe) in this pull request (and in the JIRA issue).

Does that make sense?

@skoczko do you agree with this evaluation?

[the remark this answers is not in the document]
bohdan-suprun: done

NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 10.0.0 2019-10-03
```
marcjohnson-kint: The [release guidelines|https://dev.folio.org/guidelines/release-procedures/#prepare-the-news-document] state that the news is not updated within each pull request, please remove this change.

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -76,6 +78,7 @@ public void loanPolicyWithFixedDate() throws
 
     //////////// create loan policy with foreign key to fdd
     JsonObject loanPolicyRequest = new LoanPolicyRequestBuilder()
+      .defaultPolicy()
```
marcjohnson-kint: Why is the default policy not expressed in the default constructor of the builder, rather than a separate method?

NEWS.md
bohdan-suprun: done

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
bohdan-suprun: the defaultPolicy() method constructs a rolling policy with default rolling-specific fields (periods) set. 

And in case we need a fixed policy, we will have to reset those rolling specific fields first, otherwise we will get a validation error.

marcjohnson-kint: I think I'm confused. My question was not about what the definition of the default policy is, what I'm curious about is why make clients of the `LoanPolicyRequestBuilder` have to explicitly ask for that, rather than just define it in the constructor?

What happens if the `defaultPolicy` method is not called on the builder, what policy is defined?

bohdan-suprun: So, we have two types of policies - fixed and rolling. Each type has different set of properties (periods vs FDDS).

If we construct rolling policy in the default builder's constructor, then in order to build a fixed policy we will have to 1st reset all rolling specific fields, and 2nd set the fields that we're interesting in.

Because of this, the default constructor sets values to the common fields only - name, description and ID but Fixed and Rolling specific fields keeps unassigned.

marcjohnson-kint: > So, we have two types of policies - fixed and rolling. Each type has different set of properties (periods vs FDDS).

That's a fair point. In which case I suggest we change the name of the method to `defaultRollingPolicy` or something like that.

> If we construct rolling policy in the default builder's constructor, then in order to build a fixed policy we will have to 1st reset all rolling specific fields, and 2nd set the fields that we're interesting in.

If there is no single sensible default, then I suggest making that constructor private (or getting rid of it) and converting the `defaultPolicy` to a static factory method (and renamed as above).

> Because of this, the default constructor sets values to the common fields only - name, description and ID but Fixed and Rolling specific fields keeps unassigned.

That suggests there is no need for the default constructor

bohdan-suprun: done

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -727,6 +734,66 @@ public void canDeleteALoanPolicy()
       getResponse.getStatusCode(), is(HttpURLConnection.HTTP_NOT_FOUND));
   }
 
+  @Test
+  public void canNotUseHoldAlternateRenewalLoanPeriodForFixedProfile() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotUseHoldAlternateRenewalLoanPeriodForFixedProfile() throws Exception {
```

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -727,6 +734,66 @@ public void canDeleteALoanPolicy()
...
+    JsonResponse postResponse = createCompleted.get(5, TimeUnit.SECONDS);
+
+    assertThat(postResponse.getStatusCode(), is(422));
+    assertThat(postResponse.getJson().getJsonArray("errors")
+        .getJsonObject(0).getString("message"),
+      is("Alternate Renewal Loan Period for Holds is not allowed for policies with Fixed profile"));
+  }
+
+  @Test
+  public void canNotUseRenewalsPeriodForFixedProfile() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotUseRenewalsPeriodForFixedProfile() throws Exception {
```

bohdan-suprun: done

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
bohdan-suprun: done

PR #208, opened 2019-10-08
increase default memory limits in MD

funkymalc:
- increase JVM max heap default from 256mb to 384mb.  
- Increase container limit from 384mb to 512mb.


mtraneis:
@funkymalc Would be great to know what kind of testing was done to lead to OOM, or what workflow triggered it, or if it grew over time then there must be a memory leak somewhere.

PR #209, opened 2019-10-10
CIRCSTORE-160: Create default overdue fines policy.

VRohach:
JIRA ticket: https://issues.folio.org/browse/CIRCSTORE-160.


VRohach:
@marcjohnson-kint  Could you please review this approach regarding adding new overdue-fine-policy to the system? I think this PR will create a new one and fix an impact on the circulation rules editor regarding CIRC-495.

PR #212, opened 2019-10-11
CIRCSTORE-163 Request status not changing to "Closed - Pickup expired" whe…

kostyantyn-kh:
…n Hold shelf expiration date passes

### DIFF
--- a/src/main/java/org/folio/support/ExpirationTool.java
+++ b/src/main/java/org/folio/support/ExpirationTool.java
@@ -1,6 +1,8 @@
 package org.folio.support;
 
 import static io.vertx.core.Future.succeededFuture;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
@@ -39,6 +41,8 @@ import org.folio.rest.persist.PostgresClient;
 
 public class ExpirationTool {
 
+  private static final Logger log = LoggerFactory.getLogger(ExpirationTool.class);
+
   private ExpirationTool() {
     //do nothing
   }
@@ -69,7 +73,10 @@ public class ExpirationTool {
       .compose(requests -> reorderRequests(conn, vertx, tenant, requests))
       .setHandler(v -> {
         if (v.failed()) {
-          pgClient.rollbackTx(conn, done -> future.fail(v.cause()));
+          pgClient.rollbackTx(conn, done -> {
+            log.error("Error in request processing", v.cause());
+            future.fail(v.cause());
+          });
         } else {
           pgClient.endTx(conn, done -> future.complete());
         }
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -110,6 +107,7 @@ public void canExpireAnFirstAwaitingPickupRequest()
     UUID id1 = UUID.randomUUID();
     UUID id2 = UUID.randomUUID();
     UUID id3 = UUID.randomUUID();
+    UUID id4 = UUID.randomUUID();
```
mreno-EBSCO: This new UUID doesn't seem to be used.

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -191,14 +189,14 @@ public void canExpireAnFirstOpenUnfilledRequest()
...
-    /* Status "Open - Awaiting pickup" and request expiration date in the past - should NOT be expired */
+    /* Status "Open - In transit" and request expiration date in the past - should NOT be expired */
     createEntity(
       new RequestRequestBuilder()
       .hold()
       .withId(id1_2)
       .withItemId(itemId1)
       .withPosition(2)
-      .withStatus(OPEN_NOT_YET_FILLED)
+      .withStatus(OPEN_IN_TRANSIT)
```
mreno-EBSCO: Just curious, does changing to this status demonstrate the problem with the old code and now works with the new code?

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -69,7 +73,10 @@ private ExpirationTool() {
       .compose(requests -> reorderRequests(conn, vertx, tenant, requests))
       .setHandler(v -> {
         if (v.failed()) {
-          pgClient.rollbackTx(conn, done -> future.fail(v.cause()));
+          pgClient.rollbackTx(conn, done -> {
+            log.error("Error in request processing", v.cause());
```
mreno-EBSCO: Thanks for this logging! It will help in the future.

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
kostyantyn-kh: No, unfortunately i wasn't able to reproduce scenario described in your comment
Add "In transit" requests to ensure it works as expected.

src/main/java/org/folio/support/ExpirationTool.java
kostyantyn-kh: i hope so, my bad logger wasn't added earlier

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
kostyantyn-kh: removed

PR #213, opened 2019-10-11
CIRCSTORE-164: Implement request batch update API.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-164: Create new request batch update endpoint.

### Approach:

Implemented `POST` `/request-storage-batch/requests` endpoint that allows to execute request updates **sequentially** within the **same transaction**. Any failures occurred during update will cause transaction to be rejected.

### Sample request to swap positions for two requests:
```
{
    "requests": [
        {
            "id" : "rq1 id",
            // position attribute removed
            ...
        },
        {
            "id" : "rq2 id",
            // position attribute removed
            ...
        },
        {
            "id" : "rq1 id",
            "position": 2
            ...
        },
        {
            "id" : "rq2 id",
            "position": 1
            ...
        }
    ],
    // Not required here, as 'Same' is default value
    "transactionMode": "Same"
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,11 +52,6 @@
       <version>5.4.2</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>uk.com.robust-it</groupId>
-      <artifactId>cloning</artifactId>
-      <version>1.9.11</version>
-    </dependency>
   </dependencies>
 
   <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,11 @@
       <version>5.4.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>uk.com.robust-it</groupId>
+      <artifactId>cloning</artifactId>
+      <version>1.9.11</version>
+    </dependency>
   </dependencies>
 
   <properties>
--- a/ramls/request-storage-batch.raml
+++ b/ramls/request-storage-batch.raml
@@ -1,0 +1,32 @@
+#%RAML 1.0
+title: Request Storage Batch
+version: v0.1
+protocols: [ HTTP, HTTPS ]
+baseUri: http://localhost:9130
+
+documentation:
+  - title: Batch Request Storage API
+    content: <b>Batch request operations</b>
+
+types:
+  requests-batch: !include requests-batch.json
+  errors: !include raml-util/schemas/errors.schema
+
+traits:
+  validate: !include raml-util/traits/validation.raml
+
+/request-storage-batch:
+  /requests:
+    put:
+      is: [validate]
+      body:
+        application/json:
+          type: requests-batch
+      responses:
+        204:
+          description: "Update executed successfully"
+        500:
+          description: "Internal server error, e.g. due to misconfiguration"
+          body:
+            text/plain:
+              example: "Internal server error, contact administrator"
--- a/ramls/request-storage-batch.raml
+++ b/ramls/request-storage-batch.raml
@@ -17,13 +17,13 @@ traits:
 
 /request-storage-batch:
   /requests:
-    put:
+    post:
       is: [validate]
       body:
         application/json:
           type: requests-batch
       responses:
-        204:
+        201:
           description: "Update executed successfully"
         500:
           description: "Internal server error, e.g. due to misconfiguration"
--- a/ramls/requests-batch.json
+++ b/ramls/requests-batch.json
@@ -11,19 +11,10 @@
         "type": "object",
         "$ref": "request.json"
       }
-    },
-    "transactionMode": {
-      "type": "string",
-      "description": "Transaction mode to use for update operation: Same - execute all updates sequentially within the same transaction",
-      "enum": [
-        "Same"
-      ],
-      "default": "Same"
     }
   },
   "additionalProperties": false,
   "required": [
-    "requests",
-    "transactionMode"
+    "requests"
   ]
 }
--- a/ramls/requests-batch.json
+++ b/ramls/requests-batch.json
@@ -1,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Collection of requests",
+  "type": "object",
+  "properties": {
+    "requests": {
+      "description": "List of request items to update",
+      "id": "requests",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "$ref": "request.json"
+      }
+    },
+    "transactionMode": {
+      "type": "string",
+      "description": "Transaction mode to use for update operation: Same - execute all updates sequentially within the same transaction",
+      "enum": [
+        "Same"
+      ],
+      "default": "Same"
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "requests",
+    "transactionMode"
+  ]
+}
--- a/src/main/java/org/folio/rest/impl/RequestsAPI.java
+++ b/src/main/java/org/folio/rest/impl/RequestsAPI.java
@@ -3,7 +3,8 @@ package org.folio.rest.impl;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Handler;
-import org.folio.rest.jaxrs.model.Error;
+
+import org.folio.rest.impl.util.RequestsApiUtil;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.Request;
 import org.folio.rest.jaxrs.model.Requests;
@@ -14,8 +15,6 @@ import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.tools.utils.TenantTool;
 
 import javax.ws.rs.core.Response;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
 import static io.vertx.core.Future.succeededFuture;
 import static org.folio.rest.impl.Headers.TENANT_HEADER;
@@ -130,24 +129,14 @@ public class RequestsAPI implements RequestStorage {
   }
 
   private Errors samePositionInQueueError(Request request) {
-    Error error = new Error();
-
-    error.withMessage("Cannot have more than one request with the same position in the queue")
-      .withAdditionalProperty("itemId", request.getItemId())
-      .withAdditionalProperty("position", request.getPosition());
-
-    List<Error> errorList = new ArrayList<>();
-    errorList.add(error);
-
-    Errors errors = new Errors();
-    errors.setErrors(errorList);
-
-    return errors;
+    return RequestsApiUtil
+      .samePositionInQueueError(request.getItemId(), request.getPosition());
   }
 
   private boolean isSamePositionInQueueError(AsyncResult<Response> reply) {
     return reply.succeeded()
-        && reply.result().getStatus() == 400
-        && reply.result().getEntity().toString().contains("request_itemid_position_idx_unique");
+      && reply.result().getStatus() == 400
+      && RequestsApiUtil
+      .hasSamePositionConstraintViolated(reply.result().getEntity().toString());
   }
 }
--- a/src/main/java/org/folio/service/BatchResourceService.java
+++ b/src/main/java/org/folio/service/BatchResourceService.java
@@ -1,0 +1,109 @@
+package org.folio.service;
+
+import static org.folio.rest.jaxrs.model.RequestsBatch.TransactionMode;
+
+import java.util.List;
+import java.util.function.Function;
+
+import org.folio.rest.persist.PostgresClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.ext.sql.SQLConnection;
+import io.vertx.ext.sql.UpdateResult;
+
+public class BatchResourceService {
+  private static final Logger LOG = LoggerFactory.getLogger(BatchResourceService.class);
+  private static final String WHERE_CLAUSE = "WHERE id = '%s'";
+
+  private final PostgresClient postgresClient;
+  private final String tableName;
+
+  public BatchResourceService(PostgresClient postgresClient, String tableName) {
+    this.postgresClient = postgresClient;
+    this.tableName = tableName;
+  }
+
+  /**
+   * Execute batch update.
+   *
+   * @param transactionMode - Transaction mode for update:
+   *                        Same - use one transaction for all entities
+   * @param entities        - Entities to update
+   * @param getId           - Function to get ID from an entity.
+   * @param onFinishHandler - Callback.
+   * @param <T>             - Entity type.
+   * @throws UnsupportedOperationException if transaction mode is not supported.
+   */
+  public <T> void executeBatchUpdate(
+    TransactionMode transactionMode, List<T> entities, Function<T, String> getId,
+    Handler<AsyncResult<Void>> onFinishHandler) {
+
+    // Currently only one mode supported
+    if (transactionMode == TransactionMode.SAME) {
+      executeBatchUpdateInSameTransaction(entities, getId, onFinishHandler);
+    } else {
+      throw new UnsupportedOperationException(
+        "Transaction mode: [" + transactionMode + "] is not supported now");
+    }
+  }
+
+  private <T> void executeBatchUpdateInSameTransaction(
+    List<T> entities, Function<T, String> getId, Handler<AsyncResult<Void>> onFinishHandler) {
+
+    postgresClient.startTx(connectionResult -> {
+      if (connectionResult.failed()) {
+        LOG.warn("Can not start transaction", connectionResult.cause());
+        onFinishHandler.handle(Future.failedFuture(connectionResult.cause()));
+        return;
+      }
+
+      // Using this future for chaining updates
+      Future<UpdateResult> lastUpdate = Future.succeededFuture();
+      for (T entity : entities) {
+        String id = getId.apply(entity);
+        lastUpdate = lastUpdate
+          .compose(prev -> updateSingleEntity(connectionResult, id, entity));
+      }
+
+      // Handle overall update result and decide on whether to commit or rollback transaction
+      lastUpdate.setHandler(updateResult -> {
+        if (updateResult.failed()) {
+          LOG.warn("Batch update rejected", updateResult.cause());
+
+          // Rollback transaction and keep original cause.
+          postgresClient.rollbackTx(connectionResult,
+            rollback -> onFinishHandler.handle(Future.failedFuture(updateResult.cause()))
+          );
+        } else {
+          LOG.debug("Update successful, committing transaction");
+
+          postgresClient.endTx(connectionResult, onFinishHandler);
+        }
+      });
+    });
+  }
+
+  /**
+   * Executes an update for a single entity on given DB connection.
+   *
+   * @param connection - DB connection.
+   * @param id         - Entity ID to update.
+   * @param entity     - New entity state.
+   * @param <T>        - Entity type.
+   * @return Callback.
+   */
+  private <T> Future<UpdateResult> updateSingleEntity(
+    AsyncResult<SQLConnection> connection, String id, T entity) {
+
+    Future<UpdateResult> updateResultFuture = Future.future();
+
+    postgresClient.update(connection, tableName, entity, "jsonb",
+      String.format(WHERE_CLAUSE, id), false, updateResultFuture);
+
+    return updateResultFuture;
+  }
+}
--- a/src/main/java/org/folio/service/BatchResourceService.java
+++ b/src/main/java/org/folio/service/BatchResourceService.java
@@ -1,5 +1,6 @@
 package org.folio.service;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.function.Function;
 
@@ -7,9 +8,12 @@ import org.folio.rest.persist.PostgresClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Lists;
+
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.json.JsonArray;
 import io.vertx.ext.sql.SQLConnection;
 import io.vertx.ext.sql.UpdateResult;
 
@@ -18,23 +22,19 @@ public class BatchResourceService {
   private static final String WHERE_CLAUSE = "WHERE id = '%s'";
 
   private final PostgresClient postgresClient;
-  private final String tableName;
 
-  public BatchResourceService(PostgresClient postgresClient, String tableName) {
+  public BatchResourceService(PostgresClient postgresClient) {
     this.postgresClient = postgresClient;
-    this.tableName = tableName;
   }
 
   /**
    * Execute batch update in a single transaction, subsequently.
    *
-   * @param entities        - Entities to update
-   * @param getId           - Function to get ID from an entity.
+   * @param batchFactories  - Factory to create a batch update chunk.
    * @param onFinishHandler - Callback.
-   * @param <T>             - Entity type.
    */
-  public <T> void executeBatchUpdate(
-    List<T> entities, Function<T, String> getId,
+  public void executeBatchUpdate(
+    List<Function<SQLConnection, Future<UpdateResult>>> batchFactories,
     Handler<AsyncResult<Void>> onFinishHandler) {
 
     postgresClient.startTx(connectionResult -> {
@@ -44,12 +44,12 @@ public class BatchResourceService {
         return;
       }
 
+      SQLConnection connection = connectionResult.result();
+
       // Using this future for chaining updates
       Future<UpdateResult> lastUpdate = Future.succeededFuture();
-      for (T entity : entities) {
-        String id = getId.apply(entity);
-        lastUpdate = lastUpdate
-          .compose(prev -> updateSingleEntity(connectionResult, id, entity));
+      for (Function<SQLConnection, Future<UpdateResult>> factory : batchFactories) {
+        lastUpdate = lastUpdate.compose(prev -> factory.apply(connection));
       }
 
       // Handle overall update result and decide on whether to commit or rollback transaction
@@ -71,22 +71,52 @@ public class BatchResourceService {
   }
 
   /**
-   * Executes an update for a single entity on given DB connection.
+   * Creates update single entity batch function.
    *
-   * @param connection - DB connection.
-   * @param id         - Entity ID to update.
-   * @param entity     - New entity state.
-   * @param <T>        - Entity type.
-   * @return Callback.
+   * @param tableName - Table name to update.
+   * @param id        - Entity ID.
+   * @param entity    - New entity.
+   * @param <T>       - Entity type.
+   * @return Batch function which consumes SQL connection and returns future
+   * with result of the update.
    */
-  private <T> Future<UpdateResult> updateSingleEntity(
-    AsyncResult<SQLConnection> connection, String id, T entity) {
+  public <T> Function<SQLConnection, Future<UpdateResult>> updateSingleEntityBatchFactory(
+    String tableName, String id, T entity) {
 
-    Future<UpdateResult> updateResultFuture = Future.future();
+    return connection -> {
+      Future<UpdateResult> updateResultFuture = Future.future();
+      Future<SQLConnection> connectionResult = Future.succeededFuture(connection);
 
-    postgresClient.update(connection, tableName, entity, "jsonb",
-      String.format(WHERE_CLAUSE, id), false, updateResultFuture);
+      LOG.debug("Updating entity {} with id {}", entity, id);
 
-    return updateResultFuture;
+      postgresClient.update(connectionResult, tableName, entity, "jsonb",
+        String.format(WHERE_CLAUSE, id), false, updateResultFuture);
+
+      return updateResultFuture;
+    };
+  }
+
+  /**
+   * Creates execute update query with params batch function.
+   *
+   * @param query  - SQL UPDATE/INSERT/DELETE query.
+   * @param params - SQL query params.
+   * @return Batch function which consumes SQL connection and returns future
+   * with result of the update.
+   */
+  public Function<SQLConnection, Future<UpdateResult>> queryWithParamsBatchFactory(
+    String query, Collection<?> params) {
+
+    return connection -> {
+      Future<UpdateResult> updateResultFuture = Future.future();
+      LOG.debug("Executing SQL [{}], got [{}] parameters", query, params.size());
+
+      connection.updateWithParams(query,
+        new JsonArray(Lists.newArrayList(params)),
+        updateResultFuture
+      );
+
+      return updateResultFuture;
+    };
   }
 }
--- a/src/main/java/org/folio/service/BatchResourceService.java
+++ b/src/main/java/org/folio/service/BatchResourceService.java
@@ -1,7 +1,5 @@
 package org.folio.service;
 
-import static org.folio.rest.jaxrs.model.RequestsBatch.TransactionMode;
-
 import java.util.List;
 import java.util.function.Function;
 
@@ -28,31 +26,16 @@ public class BatchResourceService {
   }
 
   /**
-   * Execute batch update.
+   * Execute batch update in a single transaction, subsequently.
    *
-   * @param transactionMode - Transaction mode for update:
-   *                        Same - use one transaction for all entities
    * @param entities        - Entities to update
    * @param getId           - Function to get ID from an entity.
    * @param onFinishHandler - Callback.
    * @param <T>             - Entity type.
-   * @throws UnsupportedOperationException if transaction mode is not supported.
    */
   public <T> void executeBatchUpdate(
-    TransactionMode transactionMode, List<T> entities, Function<T, String> getId,
+    List<T> entities, Function<T, String> getId,
     Handler<AsyncResult<Void>> onFinishHandler) {
-
-    // Currently only one mode supported
-    if (transactionMode == TransactionMode.SAME) {
-      executeBatchUpdateInSameTransaction(entities, getId, onFinishHandler);
-    } else {
-      throw new UnsupportedOperationException(
-        "Transaction mode: [" + transactionMode + "] is not supported now");
-    }
-  }
-
-  private <T> void executeBatchUpdateInSameTransaction(
-    List<T> entities, Function<T, String> getId, Handler<AsyncResult<Void>> onFinishHandler) {
 
     postgresClient.startTx(connectionResult -> {
       if (connectionResult.failed()) {
--- a/src/main/java/org/folio/service/request/RequestBatchResourceService.java
+++ b/src/main/java/org/folio/service/request/RequestBatchResourceService.java
@@ -61,7 +61,8 @@ public class RequestBatchResourceService {
     List<Request> requests, Handler<AsyncResult<Void>> onFinishHandler) {
 
     LOG.debug("Removing positions for all request to go through positions constraint");
-    List<Function<SQLConnection, Future<UpdateResult>>> allBatches = new ArrayList<>();
+    List<Function<SQLConnection, Future<UpdateResult>>> allDatabaseOperations =
+      new ArrayList<>();
 
     // Remove positions for the requests before updating them
     // in order to go through item position constraint.
@@ -71,14 +72,14 @@ public class RequestBatchResourceService {
     List<Function<SQLConnection, Future<UpdateResult>>> updateRequestsBatch =
       updateRequestsBatch(requests);
 
-    allBatches.add(removePositionBatch);
-    allBatches.addAll(updateRequestsBatch);
+    allDatabaseOperations.add(removePositionBatch);
+    allDatabaseOperations.addAll(updateRequestsBatch);
 
     LOG.info("Executing batch update, total records to update [{}] (including remove positions)",
-      allBatches.size()
+      allDatabaseOperations.size()
     );
 
-    batchResourceService.executeBatchUpdate(allBatches, onFinishHandler);
+    batchResourceService.executeBatchUpdate(allDatabaseOperations, onFinishHandler);
   }
 
   private Function<SQLConnection, Future<UpdateResult>> removePositionsForRequestsBatch(
--- a/src/main/java/org/folio/service/request/RequestBatchResourceService.java
+++ b/src/main/java/org/folio/service/request/RequestBatchResourceService.java
@@ -1,28 +1,38 @@
 package org.folio.service.request;
 
+import static org.folio.rest.impl.RequestsAPI.REQUEST_TABLE;
+
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.folio.rest.jaxrs.model.Request;
+import org.folio.rest.persist.PostgresClient;
 import org.folio.service.BatchResourceService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.rits.cloning.Cloner;
-
 import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.ext.sql.SQLConnection;
+import io.vertx.ext.sql.UpdateResult;
 
 public class RequestBatchResourceService {
   private static final Logger LOG = LoggerFactory.getLogger(RequestBatchResourceService.class);
+  private static final String REMOVE_POSITIONS_SQL =
+    "UPDATE %s.%s SET jsonb = jsonb - 'position' WHERE id::text IN (%s)";
 
   private final BatchResourceService batchResourceService;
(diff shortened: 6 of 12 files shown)


ramls/request-storage-batch.raml
```diff
@@ -0,0 +1,32 @@
...
+types:
+  requests-batch: !include requests-batch.json
+  errors: !include raml-util/schemas/errors.schema
+
+traits:
+  validate: !include raml-util/traits/validation.raml
+
+/request-storage-batch:
+  /requests:
+    put:
```
marcjohnson-kint: The experimental `instance-storage-batch` interface uses `POST` for the batch operation. Why have you chosen `PUT` instead?

ramls/requests-batch.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "transactionMode": {
+      "type": "string",
+      "description": "Transaction mode to use for update operation: Same - execute all updates sequentially within the same transaction",
+      "enum": [
+        "Same"
+      ],
+      "default": "Same"
+    }
+  },
```
marcjohnson-kint: What is the transaction mode?

If there is only a single mode, what value does introducing the parameter have?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    JsonObject firstRequestWithNoPosition = firstRequest.copy();
+    JsonObject secondRequestWithNoPosition = secondRequest.copy();
+
+    // 2.1 - release previous positions
+    secondRequestWithNoPosition.remove("position");
+    firstRequestWithNoPosition.remove("position");
+
+    // 2.2 - set new positions
+    firstRequest.put("position", 2);
+    secondRequest.put("position", 1);
```
marcjohnson-kint: Why do the positions need removing? Does the `JsonObject.put` method not overwrite an existing property value?

What do the numbers in the comments refer to? 

Could the comments be replaced with extracted methods, if there are concerns about the individual steps not conveying the intent clearly enough?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    CompletableFuture<Response> putCompleted = new CompletableFuture<>();
+    JsonObject allRequests = buildBatchUpdateRequest(firstRequestWithNoPosition,
+      secondRequestWithNoPosition, firstRequest, secondRequest);
+
+    client.put(batchRequestStorageUrl(), allRequests, TENANT_ID, empty(putCompleted));
+
+    Response response = putCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(response.getStatusCode(), is(204));
+
+    // 3rd: Retrieve requests from DB and assert changes.
```
marcjohnson-kint: What information does this comment add over the code below it?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    JsonObject firstRequestFromDb = requestsFromDb.getJsonArray("requests").getJsonObject(0);
+    JsonObject secondRequestFromDb = requestsFromDb.getJsonArray("requests").getJsonObject(1);
+
+    assertThat(requestsFromDb.getInteger("totalRecords"), is(2));
+    assertThat(firstRequestFromDb, is(firstRequest));
+    assertThat(secondRequestFromDb, is(secondRequest));
+  }
+
+  @Test
+  public void canRollbackTransactionOnSamePositionConstraintViolation() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void willAbortBatchUpdateForRequestsAtTheSamePositionInAnItemsQueue() throws Exception {
```

marcjohnson-kint: I think it could be useful to have examples/tests where the requests that conflict are not both in the batch update.

For example, introduce requestA and requestB in positions 1 and 2 respectively. The update requestB to position 2 in a batch and confirm that the action is not performed.

And a test for when there are requests for different items in the same batch, and they don't cause the operation to be aborted.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    JsonObject firstRequestCopy = firstRequest.copy();
+
+    firstRequestCopy.remove("id");
+    secondRequest.put("position", 10);
+
+    CompletableFuture<Response> putCompleted = new CompletableFuture<>();
+    client.put(batchRequestStorageUrl(),
+      buildBatchUpdateRequest(firstRequestCopy, secondRequest),
+      TENANT_ID,
+      empty(putCompleted)
```
marcjohnson-kint: What does the empty method do?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    );
+
+    // Expect server error due to Exception
+    Response response = putCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(response.getStatusCode(), is(500));
```
marcjohnson-kint: What kind of exception is thrown in this circumstance?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,192 @@
...
+    // 3rd: Retrieve requests from DB and assert that changes have not been applied.
+    JsonObject requestsFromDb = getAllRequestsForItem(itemId);
+
+    JsonObject firstRequestFromDb = requestsFromDb.getJsonArray("requests").getJsonObject(0);
+    JsonObject secondRequestFromDb = requestsFromDb.getJsonArray("requests").getJsonObject(1);
+
+    assertThat(requestsFromDb.getInteger("totalRecords"), is(2));
+    assertThat(firstRequestFromDb, is(firstRequest));
+    // 2nd RQ must have it's initial position - 2nd
+    assertThat(secondRequestFromDb, is(secondRequest.put("position", 2)));
```
marcjohnson-kint: I think it is clearer to get the position from the JSON and assert against the raw property, rather than assert against a mutated copy of entire JSON. 

Why is the entire JSON is being asserted on?

ramls/requests-batch.json
bohdan-suprun: Currently all updates are executed subsequently within the same transaction, this is what transactionMode = same defines. But it might be a case when we wont to execute updates concurrently and within separate transaction (this is how batch API for instances works now). For this we can introduce a new transactionMode. 

I've added this parameter to set up a basic architecture of the resource.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: > Why do the positions need removing? Does the JsonObject.put method not overwrite an existing property value?

We have to release the position before assigning it to another request. This is because position constraint are being evaluated after each row update.

> What do the numbers in the comments refer to?

They refers to a flow step.

> Could the comments be replaced with extracted methods, if there are concerns about the individual steps not conveying the intent clearly enough?

For me, it increases code readability when we have such comments, so I can understand what this test do without diving into code.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: it is async handler for a request without body.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: It is NPE, because Id is removed but it is required to for update in where clause.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: The main purpose of asserting entire JSON to make sure that not even position has not been changed but all other fields are not modified.

ramls/request-storage-batch.raml
bohdan-suprun: The `instance-storage-batch` is used to create instances in batch, but this endpoint is going to be used for updates. 

Do you think it should be POST? And which method will we use if we decide to implement endpoint to save requests in batch?

ramls/requests-batch.json
marcjohnson-kint: > Currently all updates are executed subsequently within the same transaction, this is what transactionMode = same defines. But it might be a case when we wont to execute updates concurrently and within separate transaction (this is how batch API for instances works now). For this we can introduce a new transactionMode.
> 
> I've added this parameter to set up a basic architecture of the resource.

I can understand creating space for these kinds of options, by using an object and having the requests as a property.

What I don't understand is why commit to the details of the parameters at this stage, when they don't offer any choices for the client? 

Maybe it is trying to guide the future changes in a certain way? I'm cautious about that because it can lead to premature commitment. I understand that deferring can sometimes lead to misunderstandings when it is introduced.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: > > Why do the positions need removing? Does the JsonObject.put method not overwrite an existing property value?
> 
> We have to release the position before assigning it to another request. This is because position constraint are being evaluated after each row update.

Why do we have to release the position, I thought the intent of this batching was that clients no longer had to do that?

> > What do the numbers in the comments refer to?
> 
> They refers to a flow step.

What information does a flow step provide?

> > Could the comments be replaced with extracted methods, if there are concerns about the individual steps not conveying the intent clearly enough?
> 
> For me, it increases code readability when we have such comments, so I can understand what this test do without diving into code.

A big tradeoff for comments from my perspective is that they can easily get out of sync. I'd prefer extracting of well named methods that describe the flow at a higher level.

bohdan-suprun: > Why do we have to release the position, I thought the intent of this batching was that clients no longer had to do that?

No, we still have to release them, but it is more safe to failures during update: If there is a failure during position release/update all the changes will be reverted to initial state (as they were before reorder).

Unfortunately, there is no way to bypass the constraint, as by default unique constraints check are executed after each row update (not during transaction commit).

mreno-EBSCO: Wait, what? I don't follow your reasoning here. Why can't we just remove the positions as part of the transaction? Why do we need to explicitly remove them as part of the input batch? Are we trying to avoid parsing the input JSON? I think it is pretty heavy handed to force the client to do this. We should be able to solve this all in the backend. The constraints are the storage module's problem. In fact, I would even say that it should be a validation error to pass in a position that is `null` or not set.

Maybe I am missing something here...

bohdan-suprun: @mreno-EBSCO, yep, make sense. But in this case the batch endpoint will serve only one particular case - request position update, that for me is a responsibility of the business module, and I'm not sure that it is fine to introduce such endpoint in storage module. 

> We should be able to solve this all in the backend. 

I don't think that this endpoint will be executed directly from UI, but as a part of an business flow.

What do you think?

mreno-EBSCO: > But in this case the batch endpoint will serve only one particular case - request position update, that for me is a responsibility of the business module, and I'm not sure that it is fine to introduce such endpoint in storage module.

But isn't that exactly what this API is for, moving requests in the queue in an atomic way? I don't think it would be/should be used for anything else. Is that really what we are trying to accomplish here, a general purpose batch API to update requests? Is there a need for that?

I think we should be creating a specific POST endpoint for updating request queues of a specific item in an atomic way and leave it at that.

> I don't think that this endpoint will be executed directly from UI, but as a part of an business flow.

These are public endpoints. Any client with the correct permissions can call them and do whatever they are allowed to by the API, potentially updating every request in this case. This is one of the reasons why I think we need a very specific endpoint for request queue reordering.

I thought we were going to have a simple API, perhaps based on an itemId, where we pass in JSON that is just an array of objects containing the request ID and the new position, not the complete requests, and the storage module would do the reordering in a transaction. A general purpose batch API will require so much validation to ensure nothing is being corrupted. I just don't see the feasibility here.

I must be missing something...

bohdan-suprun: @mreno-EBSCO, I totally agree with you're statements above.

@marcjohnson-kint, are you ok with this approach? Introduce an endpoint in storage module that will supply the same reorderedQueue as business service and do update of positions only.

mreno-EBSCO: Oh, I see. We are going to do a bunch of this work in `mod-circulation`. I still think this should not be a general purpose batch API.

marcjohnson-kint: @bohdan-suprun I'm ok with a dedicated change positions endpoint, as long as it is in a separate interface, and I would like keep it experimental.

A concern I have is that it might appear similar to the higher level business logic endpoint, and that might be confusing. We want folks to use the business logic API, yet we cannot stop them using the storage API (I can refer folks to my comments elsewhere about some of the challenges with this)

Alas, as @mreno-EBSCO stated above, these are public APIs these kinds of aspects are inherent in the storage and business logic split.

mreno-EBSCO: I don't want to change scope here, so maybe my comments are not appropriate.

That being said... Do we need an endpoint in the BL module for this? If so, couldn't it just be a pass through? I think anything we do in the BL module, we'd need to do in the storage module.

marcjohnson-kint: I think we need do consider how much logic should be in storage, for example, should the storage API include validation for contiguous positions? Or that requests in certain statuses or types should not move from position 1?

mreno-EBSCO: > ... should the storage API include validation for contiguous positions? Or that requests in certain statuses or types should not move from position 1?

Ah, good point.

I am always nervous about the faith we have in clients not to call storage APIs with inconsistent but valid data.

bohdan-suprun: I think it might be some issues with moving validation to storage module. Because in different cases we can allow or disallow same things.

For example for queue reorder we don't allow requests with started fulfillment to be moved from the 1st position. But this might be allowed when we chek-in an item and shift all requests in the queue. 

But I may be wrong.

mreno-EBSCO: Yeah, I agree, there are challenges here given how much business logic is required to reorder the queue and the need to make this update atomic. My initial comments were a bit off, since I forgot the simpler API would be in mod-circulation.

However, I am concerned with how powerful the batch API is and that we might get in trouble with end users using it instead of mod-circulation thinking it is how they could bulk load request updates into the system. Then again, I'm not sure if that is something anyone would want to do.

marcjohnson-kint: @mreno-EBSCO As I understand it, FOLIO as an architecture has decided to introduce batch APIs for both storage and business logic, so they are likely coming.

What I'm most concerned about is that we get a thorough understanding and agreement on the behaviour and design of that before we build too many of them. Which is why all of the existing ones are hopefully marked experimental and kept separate.

I think the topic of storage vs. business logic API use is an increasingly important and broad one, especially as folks start to build migrations tooling. I'm going to add it to the tech leads topic list.

@mreno-EBSCO @bohdan-suprun I think we need to decide whether to introduce a specific position only storage API or a very rough and experimental general batch API?

If we want to support update of the queue during fulfilment, then we likely need a more general API.

marcjohnson-kint: On the subject of removing the positions first. I have two questions:

Why do we need to express this separate step in the tests, as I understand it, that is only relevant in the implementation?

Can we use [deferrable constraints](https://hashrocket.com/blog/posts/deferring-database-constraints) to avoid this in the implementation as well?

bohdan-suprun: Yep, deferrable constraints is a good option, but it seems RMB does not support it out of the box, so we will need to add it in a separate script  + care about migration (remove old constraint and replace it with the new one).

> Why do we need to express this separate step in the tests, as I understand it, that is only relevant in the implementation?

Well, currently the API assumes that client better know how to execute particular update (i mean order of updates, dealing with constraints, etc.) to get it successfully committed.

ramls/request-storage-batch.raml
marcjohnson-kint: Similarly to the conversation about the reordering API, I think Thor API is better suited to POST, because it is really a request to process a batch, rather than a create or replace of a specific resource.

It also keeps it similar to the other experimental batch APIs.

@mreno-EBSCO what do you think?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: @bohdan-suprun thanks for investigating the deferred constraints

marcjohnson-kint: I can understand that. Maybe there is a way to make that clearer, maybe by not doing the position mutation inline, or using the original representation prior to the attempted operation?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: Please can the test name be updated to describe that better

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: @bohdan-suprun thanks for investigating the deferred constraints

bohdan-suprun: Hmm... I think we will be able to detect the fact that we're going to change positions in the batch, but this will require additional query to db (so we will compare provided positions vs positions in DB). And further, we can add a custom logic for this particular case.

@marcjohnson-kint, @mreno-EBSCO, what do you think?

marcjohnson-kint: bq. The client have to take care about satisfying all the constraints (as in this case release all positions before assigning to pass item-position constraint). In a fact client should be aware about DB constraints, data consistency. etc.

If we are expecting the client to release the positions and then re-specify the new ones as two separate operations, what are we gaining by introducing this new storage API?

I think it is reasonable that the client satisfy the expectations that positions should be unique, I think we are leaking too much for a new batch API if we expect them to understand they need to do this in two operations.

I had thought we were trying to hide this from the client (in this case, mod-circulation).

> Hmm... I think we will be able to detect the fact that we're going to change positions in the batch, but this will require additional query to db (so we will compare provided positions vs positions in DB). And further, we can add a custom logic for this particular case.

I'm starting to understand how this could be complicated, because the batch API could be used for a mixture of requests for different items, or for only parts of a queue.

I'm beginning to wonder if we are ready to try to design a better storage API to support atomic updates with these constraint limitations?

bohdan-suprun: > If we are expecting the client to release the positions and then re-specify the new ones as two separate operations, what are we gaining by introducing this new storage API?

Well, we don't need two operations actually. We still need one operation, but batch package have to include requests where positions are removed and requests with the new positions set.

> I'm beginning to wonder if we are ready to try to design a better storage API to support atomic updates with these constraint limitations?

Maybe not, but from my point of view it more reliably then sending those updates concurrently in different transactions.

marcjohnson-kint: bq. Well, we don't need two operations actually. We still need one operation, but batch package have to include requests where positions are removed and requests with the new positions set.

I think I need an example. Let's say we have a request queue for Harry Potter and the Cursed Child:
Request A, Position 1
Request B, Position 2
Request C, Position 3

Can the batch API be used, with a single operation, to perform the following reorder?
Request A, Position 1
Request C, Position 2
Request B, Position 3

bohdan-suprun: Yep, the batch API will look like follow:

Request A: with no position 
Request C: with no position 
Request B: with no position 
Request A, Position 1
Request C, Position 2
Request B, Position 3

As you can see, It includes 3 additional copies of the requests with positions removed. They must be executed before setting new positions.

marcjohnson-kint: Hang on, we are expecting a client to provide two representations of the same record within the same batch API request?

I really don't think we want to expose that complexity to the clients for a new API

bohdan-suprun: Then the only option to implement API that will update positions only.

marcjohnson-kint: If we accept this limitation in the current batch API, which I think is ok. Then I think we should do it in two operations.

The first to clear the positions, the second to update them. I know that isn't ideal, as it isn't atomic. I think it is simpler to understand than this kind of request.

@mreno-EBSCO  @mtraneis @jsavell what do you think?

marcjohnson-kint: I think we might want to take a step back and decide if we really want to invest more in this right now, given the complications of the implementation

bohdan-suprun: > If we accept this limitation in the current batch API, which I think is ok. Then I think we should do it in two operations.

Executing it as two operations won't give us the main benefit of the batching. As @mreno-EBSCO mentioned in https://github.com/folio-org/mod-circulation/pull/366, it will be not possible to recover in case when removal executed successfully but position update has been failed. That's really the main reason to use this API (at least for the queue reorder feature).

marcjohnson-kint: Then I think we should either abort this work and accept the current limitations, that failures can lead to a corrupted queue, or build a dedicated reordering API in storage with very limited behaviour.

I don’t think we are ready for a general purpose batch API, and submitting two representations of the same record and relying on it doing separate updates in order feels like it could be really brittle to me.

@mreno-EBSCO @skoczko what do you think?

I wonder if we’ve already invested more time than we really had in this

mreno-EBSCO: So, what harm is done if we simply remove the `position` field of all the requests in the batch from what is stored in the DB prior to writing the batch? We do this when requests are expired. If we are assuming that the client is doing the right thing prior to calling the API then the existing positions should not matter. If they do, then the client missed something or the queue was modified after the client received the list, but before it sent the request and likely this will lead to a DB constraint violation and a transaction rollback.

Anyway, removing the positions as the first part of the transaction shouldn't matter even in a general purpose batch API. I don't think this is business logic, so much as requirements to write to the DB in this storage module. I don't think clients need to know this, nor should they, since there is no real way to convey this (maybe the API doc, but that isn't ideal). Even if positions are not changed, they could still be removed, since the request object being written will have the correct position already assigned, if not DB rollback, no harm...

mreno-EBSCO: @marc-johnson Corrupting the queue leads to problems elsewhere and is confusing in the UI. We've already had to fix queues several times for Chalmers. Manually fixing queues will not scale. We cannot corrupt the queue. Chalmers gave an item to a patron without a loan because the patron was not next in the request queue, as they should have been, because of a corrupt queue. So, I do think we need to address this with some sort of transaction based reordering. Like I said above, I think if we just remove the `position` field for each request in the batch as the first part of the transaction, then I think we'll be OK. What do you think about that approach?

ramls/request-storage-batch.raml
bohdan-suprun: Modified to POST now.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: I've refactored tests, it is outdated now.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: I've added these test cases.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: Updated.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: Outdated.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: @mreno-EBSCO i nearly raised the idea of removing the position internally to the batch API this morning.

I aborted as I started thinking of the various ways it could be possible for the API request to fail, and then got involved in other things.

I think we need to clearly document this behaviour. I agree it is much cleaner than submitting the records twice.

Let’s try it, I can’t remember why I shied away from it this morning.

bohdan-suprun: Hi @marcjohnson-kint,

Are we proceeding with the option when the batch API will detect whether positions has been changed or not and then add request representations with removed positions before updating? 

Another option that we have, if we agreed that supplying two representations of the same request is unacceptable, is following. 

Get rid of json-based unique index, but create two table columns instead (`itemId` and `position`) which will be populated by a trigger. So then we can create `deferrable` constraint and make things much easier. Is it make sense?

marcjohnson-kint: @bohdan-suprun What do you think of the idea that the batch API removes the positions from all requests submitted, as a first step, and then does the regular updates, all within the implementation?

My understanding of that approach would be something like:

Let's say we have a request queue for Harry Potter and the Cursed Child:
Request A, Position 1
Request B, Position 2
Request C, Position 3

When the following requests are submitted to the batch API
Request A, Position 1
Request C, Position 2
Request B, Position 3

Internally the steps are as follows:

1. Start transaction
2. Remove the positions from all requests involved in the batch
3. Update the requests as usual
4. Commit transaction

This will in effect be an imitation of deferred constraint.

I think this is much cleaner and better reflects the general purpose nature of the API.

I had reservations about suggesting that yesterday, as I started to get concerned about what happens if the API is used for partial queue updates or requests for multiple items, but can't seem to think of specific issues, as every request whose position changes should be involved in the batch submission.

I think we need to make sure we document the reasons behind this and the limitations.

marcjohnson-kint: I think if that approach doesn't work well, then I think we should go with the approach you suggested and swap the expression based constraint for triggers and a regular, deferrable constraint, even though I'm not a fan of triggers.

bohdan-suprun: > What do you think of the idea that the batch API removes the positions from all requests submitted, as a first step, and then does the regular updates, all within the implementation?

I can not find any issues with such approach. I'm fine with this.

bohdan-suprun: I've committed the changes that you've suggested. I have couple of questions:
1. Jenkins says following in build logs:
```
WARNING: lint-raml-cop: Missing from configuration: request-storage-batch.raml
WARNING: lint-raml-cop: Configuration needs to be updated (FOLIO-903).
```
Should I create a PR to the configuration repo?

2. 
> I think we need to make sure we document the reasons behind this and the limitations.

Do we have a separate documentation somewhere on wiki? Or you mean Java docs?

marcjohnson-kint: > Should I create a PR to the configuration repo?

I think @dcrossleyau is best placed to answer that question. It can be tested locally, however the global configuration should only be changed after this is merged. As once it is added, the documentation build will expect that RAML file to be present.

> Do we have a separate documentation somewhere on wiki? Or you mean Java docs?

I am referring to the Java docs or the readme / documentation within the repository. I don't think implementation specific information should be kept on the wiki.

pom.xml
```diff
@@ -52,6 +52,11 @@
       <version>5.4.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>uk.com.robust-it</groupId>
+      <artifactId>cloning</artifactId>
+      <version>1.9.11</version>
+    </dependency>
```
marcjohnson-kint: What is this dependency used for?

src/main/java/org/folio/rest/impl/RequestsAPI.java
```diff
@@ -130,24 +129,14 @@ public void putRequestStorageRequestsByRequestId(
...
-    Errors errors = new Errors();
-    errors.setErrors(errorList);
-
-    return errors;
+    return RequestsApiUtil
+      .samePositionInQueueError(request.getItemId(), request.getPosition());
   }
 
   private boolean isSamePositionInQueueError(AsyncResult<Response> reply) {
     return reply.succeeded()
```
marcjohnson-kint: What happens if the action failed?

src/main/java/org/folio/service/request/RequestBatchResourceService.java
```diff
@@ -0,0 +1,91 @@
...
+import com.rits.cloning.Cloner;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+
+public class RequestBatchResourceService {
+  private static final Logger LOG = LoggerFactory.getLogger(RequestBatchResourceService.class);
+
+  private final BatchResourceService batchResourceService;
+  private final Cloner cloner;
```
marcjohnson-kint: Why does the cloner need to be a field? It appears it is only used within the `cloneRequest` method.

pom.xml
bohdan-suprun: It is used to make a `Request` copy (so it will be added as the first part of transaction with no position)

src/main/java/org/folio/rest/impl/RequestsAPI.java
bohdan-suprun: I think the `isSamePositionInQueueError` method will return `false` in this case.

I did not modify original logic of this method, I just replaced `reply.result().getEntity().toString().contains("request_itemid_position_idx_unique");` with `RequestsApiUtil.hasSamePositionConstraintViolated(reply.result().getEntity().toString())`, so I can reuse it in RequestsBatchApi.

src/main/java/org/folio/service/request/RequestBatchResourceService.java
bohdan-suprun: Id does not have a state, so the same instance can be used to clone all the requests in a batch package.

pom.xml
mreno-EBSCO: I looked at this dependency and I don't think it is appropriate. There are many issues filed that indicate the library does not work with Java 10+ unless you start Java with an argument to allow reflection to work outside the module barriers put in place in the latest Java. We will likely be going to Java 11 sometime next year ([FOLIO-1772](https://issues.folio.org/browse/FOLIO-1772)). Perhaps they will solve this problem by that time, but, given their current lack of attention to the existing issues highlighting the failure, it does not seem likely.

src/main/java/org/folio/service/request/RequestBatchResourceService.java
```diff
@@ -0,0 +1,91 @@
...
+    List<Request> requestsWithRemovedPositions = removePositionsForRequests(requests);
+
+    List<Request> allRequestsToUpdate = new ArrayList<>(requestsWithRemovedPositions);
+    allRequestsToUpdate.addAll(requests);
+
+    LOG.info("Executing batch update, total records to update [{}] (including remove positions)",
+      allRequestsToUpdate.size()
+    );
+
+    batchResourceService.executeBatchUpdate(allRequestsToUpdate,
```
mreno-EBSCO: Instead of doing this cloning with `null` positions, can we just pass in some pre-batch SQL that removes the position for each request ID, like this code:
https://github.com/folio-org/mod-circulation-storage/blob/fa913ecea1fdd9af14dc69063b5c1e533db2b9a4/src/main/java/org/folio/support/ExpirationTool.java#L235-L239

It think pre/post-batch SQL might be good options for the batch executor in order to do other stuff within the batch transaction.

pom.xml
marcjohnson-kint: Is the [Apache Commons SerializationUtils.clone](http://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/SerializationUtils.html) method appropriate?

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -0,0 +1,314 @@
...
+    JsonObject allRequestsForItem = getAllRequestsForItem(itemId);
+    assertThat(allRequestsForItem.getInteger("totalRecords"), is(2));
+
+    JsonArray allRequests = allRequestsForItem.getJsonArray("requests");
+    assertThat(firstRequest, is(allRequests.getJsonObject(0)));
+    assertThat(secondRequest, is(allRequests.getJsonObject(1)));
+  }
+
+  @Test
+  public void canUpdateRequestFulfilmentPreferenceInBatch() throws Exception {
```
marcjohnson-kint: Why would the fulfilment preferences need to be updated in a batch?

src/main/java/org/folio/service/BatchResourceService.java
```diff
@@ -0,0 +1,92 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.ext.sql.SQLConnection;
+import io.vertx.ext.sql.UpdateResult;
+
+public class BatchResourceService {
```
marcjohnson-kint: Has this class been copied from elsewhere?

src/main/java/org/folio/service/request/RequestBatchResourceService.java
```diff
@@ -0,0 +1,91 @@
...
+import org.folio.service.BatchResourceService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.rits.cloning.Cloner;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+
+public class RequestBatchResourceService {
```
marcjohnson-kint: This class is named liked it is a specialisation of the `BatchResourceService` yet they have a composition relationship (which I prefer).

I'd like to understand better why these two classes are split?

src/main/java/org/folio/service/request/RequestBatchResourceService.java
marcjohnson-kint: Is it particularly expensive to instantiate a cloner? I'm asking because separating it into a field that needs initialising in the constructor means knowing about 3 places rather than only one.

src/main/java/org/folio/rest/impl/RequestsAPI.java
marcjohnson-kint: Ok, if a failure will trigger a 500 error message, then I'm ok with that. Thank you for sharing an explanation with me.

src/main/java/org/folio/service/request/RequestBatchResourceService.java
marcjohnson-kint: I prefer the SQL based approach over the deep copy and an update statement per record.

pom.xml
marcjohnson-kint: [This article](https://www.baeldung.com/java-deep-copy) suggested doing a roundtrip using Jackson to achieve a deep copy.

If we decide to continue down this route, I suggest we use that mechanism for the time being. 

I am aware it is not ideal, as it requires handling checked exceptions and likely has a performance impact.

A possible implementation provided below.

```
    final ObjectMapper mapper = ObjectMapperTool.getMapper();

    return mapper.readValue(mapper.writeValueAsString(requestToCopy), Request.class);
```

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
dcrossleyau: Thanks Marc. Yes the warnings about extra configuration can be ignored.

All RAML files are still automatically discovered and processed.

Also after merge to master, all API docs (including new ones) will still be generated and published. But as explained in the configuration documentation, they will just not be linked from the [API docs](https://dev.folio.org/reference/api/) table.

We could adjust the configuration for every new RAML file added via a branch. However then there would be a warning message in the opposite direction for the other branches, about configuration not matching the discovered set of files. Note that i manually follow up each day, to add new configuration for all master branches.

By the way, all of these warning and info messages would all be gone from the PR screen, if people attended to the missing schema descriptions.

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
bohdan-suprun: to make sure that we still able to update different from `position` property without issues.

src/main/java/org/folio/service/BatchResourceService.java
bohdan-suprun: nope

src/main/java/org/folio/service/request/RequestBatchResourceService.java
bohdan-suprun: `BatchResourceService` contains common methods for batch updates, but `RequestBatchResourceService` is implementation for requests batch updates and it contains it's specific logic (e.g. removing positions before assigning new).

pom.xml
bohdan-suprun: I think it is a common issue with reflection API since Java 9. It should affect even Mockito and PowerMockito libs. Anyway, I've chosen to use SQL query to remove positions instead of copy.

mreno-EBSCO: The difference is those are test frameworks, where we could loosen restrictions if need be. This was in production code. Anyway, the dependency is no longer needed, so I am fine. 😄

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
marcjohnson-kint: Ok. thanks for explaining that to me

src/main/java/org/folio/service/request/RequestBatchResourceService.java
```diff
@@ -0,0 +1,113 @@
...
+   * 6. Request C.
+   *
+   * @param requests        - List of requests to execute in batch.
+   * @param onFinishHandler - Callback function.
+   */
+  public void executeRequestBatchUpdate(
+    List<Request> requests, Handler<AsyncResult<Void>> onFinishHandler) {
+
+    LOG.debug("Removing positions for all request to go through positions constraint");
+    List<Function<SQLConnection, Future<UpdateResult>>> allBatches = new ArrayList<>();
```
marcjohnson-kint: Are these really batches, or are they a sequence of SQL commands?

src/main/java/org/folio/service/BatchResourceService.java
```diff
@@ -0,0 +1,122 @@
...
+   * Creates update single entity batch function.
+   *
+   * @param tableName - Table name to update.
+   * @param id        - Entity ID.
+   * @param entity    - New entity.
+   * @param <T>       - Entity type.
+   * @return Batch function which consumes SQL connection and returns future
+   * with result of the update.
+   */
+  public <T> Function<SQLConnection, Future<UpdateResult>> updateSingleEntityBatchFactory(
```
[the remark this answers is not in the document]
bohdan-suprun: This method returns a function that executes an update using connection. Sounds like a factory for me.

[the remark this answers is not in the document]
bohdan-suprun: This method returns a function that executes an update using connection. Sounds like a factory for me.

src/main/java/org/folio/service/request/RequestBatchResourceService.java
marcjohnson-kint: By the time the request updates are added to this list, are they actually SQL commands?

marcjohnson-kint: Maybe we need a different name, e.g. database operations?

src/main/java/org/folio/service/BatchResourceService.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok, personally I don't think the factory terminology adds anything to the description of the method. I appreciate your explanation, and will resolve my comment.

src/main/java/org/folio/service/request/RequestBatchResourceService.java
bohdan-suprun: > By the time the request updates are added to this list, are they actually SQL commands?

No, I use `update` method from `PostgresClient` class for update a request.

> Maybe we need a different name, e.g. all database operations?

Works for me.

bohdan-suprun:
@mreno-EBSCO, I assumed you don't have any objections/comments for this PR. Merging the PR then.

PR #214, opened 2019-10-22
CIRCSTORE-165: Rename "Hold Shelf" to be consistent with request attributes

andrii-paias:
Endpoint "/request-storage/requests" accepts/returns attribute "fulfilmentPreference" with value "Hold Shelf" (Shelf is capitalized). We need to change "/request-preference-storage/request-preference" to have capitalized "Hold Shelf", so that UI can transfer value from one endpoint to another.


andrii-paias:
@marc-johnson, 
there are unit tests - RequestPreferencesApiTest
they use java enum value to check for equality instead of String constant, so changing schema also changes values that are check in tests

bohdan-suprun:
Merging as per @andrii-paias request.

PR #218, opened 2019-11-14
CIRCSTORE-169  Add "Open - Awaiting delivery" status to Requests

dmtkachenko:
Add "Open - Awaiting delivery" status to support Fulfillment Delivery requests
Update request.json with the new status, make respective changes in the tests

### DIFF
--- a/ramls/request.json
+++ b/ramls/request.json
@@ -40,6 +40,7 @@
         "Open - Not yet filled",
         "Open - Awaiting pickup",
         "Open - In transit",
+        "Open - Awaiting delivery",
         "Closed - Filled",
         "Closed - Cancelled",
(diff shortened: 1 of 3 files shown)


marcjohnson-kint:
This pull request should have included a change to the `request-storage` interface version.

Without that, there is no way for a client to bind to the version of the API that introduces the new status value.

This is important because API requests to create or update the request status to `Open - Awaiting delivery` will work with a newer version and not with an older version, and there is no way to express that in FOLIO's dependency management

marcjohnson-kint:
To be more specific. Both the `request-storage` and the `request-storage-batch` interface definitions need a minor version increment (in both the module descriptor and the RAML files).

PR #221, opened 2019-11-18
[CIRCSTORE-170] Add Item Limit to Loan Policy

OleksandrVidinieiev:
**Purpose**
Add a new field to `Loan Policy`.

**Link**
[CIRCSTORE-170](https://issues.folio.org/browse/CIRCSTORE-170)

### DIFF
--- a/ramls/loan-policy.json
+++ b/ramls/loan-policy.json
@@ -44,6 +44,12 @@
         "fixedDueDateScheduleId": {
           "type": "string",
           "description": "Fixed due date schedule (due date limit)"
+        },
+        "itemLimit": {
+          "type": "integer",
+          "description": "Number of items allowed",
+          "minimum": 1,
+          "maximum": 9999
         }
       }
     },
--- a/src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
+++ b/src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
@@ -315,6 +315,93 @@ public class LoanPoliciesApiTest extends ApiTests {
   }
 
   @Test
+  public void canCreateLoanPolicyWithItemLimitWithinBounds()
+    throws MalformedURLException,
+    InterruptedException,
+    ExecutionException,
+    TimeoutException {
+
+    CompletableFuture<JsonResponse> createCompleted = new CompletableFuture<>();
+
+    UUID id = UUID.randomUUID();
+
+    JsonObject loanPolicyRequest = defaultRollingPolicy()
+      .withId(id)
+      .withName("Example Loan Policy")
+      .withDescription("An example loan policy")
+      .create();
+
+    loanPolicyRequest.getJsonObject("loansPolicy").put("itemLimit", 1000);
+
+    client.post(loanPolicyStorageUrl(),
+      loanPolicyRequest, StorageTestSuite.TENANT_ID,
+      ResponseHandler.json(createCompleted));
+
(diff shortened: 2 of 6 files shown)


src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -314,6 +314,93 @@ public void cannotCreateALoanPolicyWithAdditionalPropertiesInLoanPolicy()
...
+
+    loanPolicyRequest.getJsonObject("loansPolicy").put("itemLimit", 1000);
+
+    client.post(loanPolicyStorageUrl(),
+      loanPolicyRequest, StorageTestSuite.TENANT_ID,
+      ResponseHandler.json(createCompleted));
+
+    JsonResponse response = createCompleted.get(5, TimeUnit.SECONDS);
+
+    assertThat(String.format("Failed to create loan policy: %s", response.getBody()),
```
bohdan-suprun: Could you please add an assertion to check that the property is really got persisted (i.e. it returned in post response, or read the policy by ID and verify that the property is present)?

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -314,6 +314,93 @@ public void cannotCreateALoanPolicyWithAdditionalPropertiesInLoanPolicy()
...
+    loanPolicyRequest.getJsonObject("loansPolicy").put("itemLimit", 0);
+
+    client.post(loanPolicyStorageUrl(),
+      loanPolicyRequest, StorageTestSuite.TENANT_ID,
+      ResponseHandler.json(createCompleted));
+
+    JsonResponse response = createCompleted.get(5, TimeUnit.SECONDS);
+
+    assertThat(String.format("Should fail to create loan policy: %s", response.getBody()),
+      response.getStatusCode(), is(HttpStatus.HTTP_INTERNAL_SERVER_ERROR.toInt()));
```
bohdan-suprun: I think it make sense to assert error message as well, in order to make sure that the error is caused by validation rather then another error.

Hmm... another important note, why do we not expect validation error (422) in this case, but server error instead?

Could you please check if it is something related to https://issues.folio.org/browse/RMB-503 that @mreno-EBSCO observed? He has posted a temp fix for this issue.

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
OleksandrVidinieiev: done

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
OleksandrVidinieiev: Adding this missing dependency did help, thanks!

ramls/loan-policy.json
```diff
@@ -44,6 +44,12 @@
         "fixedDueDateScheduleId": {
           "type": "string",
           "description": "Fixed due date schedule (due date limit)"
+        },
+        "itemLimit": {
```
marcjohnson-kint: What is the item limit? Is it the number of items a patron can have checked out at the same time?

OleksandrVidinieiev: Hi @marcjohnson-kint, please find a detailed explanation of what an `Item limit` is in the description of [CIRC-558](https://issues.folio.org/browse/CIRC-558).

PR #224, opened 2019-11-25
Circstore 175

mrdpopov:
# Purpose

Add new property for storing the date the item was declared lost

This resolves: CIRCSTORE-175

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -15,7 +15,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "6.2",
+      "version": "6.3",
       "handlers": [
         {
           "methods": ["GET"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -15,7 +15,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "6.3",
+      "version": "6.4",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -95,7 +95,8 @@
     },
     "declaredLostDate" : {
       "description": "Date and time loan was declared lost",
-      "type": "string"
+      "type" : "string",
+      "format": "date-time"
     },
     "metadata": {
       "description": "Metadata about creation and changes to loan, provided by the server (client should not provide)",
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -93,6 +93,10 @@
       "description": "Indicates whether or not this loan had its due date modified by a recall on the loaned item",
       "type": "boolean"
     },
+    "declaredLostDate" : {
+      "description": "Date and time loan was declared lost",
+      "type": "string"
+    },
     "metadata": {
       "description": "Metadata about creation and changes to loan, provided by the server (client should not provide)",
       "type": "object",
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -94,7 +94,7 @@
       "type": "boolean"
     },
     "declaredLostDate" : {
-      "description": "Date and time loan was declared lost",
+      "description": "Date and time the item was declared lost during this loan",
       "type" : "string",
       "format": "date-time"
     },
--- a/src/test/java/org/folio/rest/api/LoansApiTest.java
+++ b/src/test/java/org/folio/rest/api/LoansApiTest.java
@@ -12,6 +12,7 @@ import static org.folio.rest.support.matchers.ValidationResponseMatchers.isValid
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsNot.not;
 import static org.hamcrest.core.IsNull.notNullValue;
@@ -137,9 +138,9 @@ public class LoansApiTest extends ApiTests {
     assertThat("recall changed due date should be null",
         loan.getBoolean("dueDateChangedByRecall"), nullValue());
 
-    assertThat("Loan should have a declaredLostDate property",
-      loansClient.getById(id).getJson().getString("declaredLostDate"),
-      is(expectedLostDate.toString(ISODateTimeFormat.dateTime())));
+    assertThat("Loan should have a declaredLostDate property", DateTime
+      .parse(loansClient.getById(id).getJson().getString("declaredLostDate"))
+      .getMillis(), is(expectedLostDate.getMillis()));
 
   }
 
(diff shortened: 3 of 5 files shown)


ramls/loan.json
```diff
@@ -93,6 +93,10 @@
       "description": "Indicates whether or not this loan had its due date modified by a recall on the loaned item",
       "type": "boolean"
     },
+    "declaredLostDate" : {
+      "description": "Date and time loan was declared lost",
+      "type": "string"
+    },
```
bohdan-suprun: do we need `"format": "date-time"` here?

descriptors/ModuleDescriptor-template.json
```diff
@@ -15,7 +15,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "6.2",
+      "version": "6.3",
```
bohdan-suprun: Please, update version in the RAML file as well.

mrdpopov: it's 6.3 already there.

ramls/loan.json
mrdpopov: Maybe we do, but all other dates are stored as strings. Do you think we should stay consistent?

mrdpopov: Sure.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Was that a typo in a previous change or has a version been missed in the module descriptor?

ramls/loan.json
```diff
@@ -93,6 +93,11 @@
       "description": "Indicates whether or not this loan had its due date modified by a recall on the loaned item",
       "type": "boolean"
     },
+    "declaredLostDate" : {
```
marcjohnson-kint: I believe there was also a discussion on the issue about a comment about why an item was declared lost, is that going to be implemented as well?

src/test/java/org/folio/rest/support/builders/LoanRequestBuilder.java
```diff
@@ -10,22 +10,23 @@
...
+  private UUID userId;
+  private UUID proxyUserId;
+  private DateTime loanDate;
+  private String statusName;
+  private String itemStatus;
+  private DateTime dueDate;
+  private String action;
+  private String actionComment;
+  private UUID itemEffectiveLocationAtCheckOut;
+  private UUID loanPolicyId;
```
marcjohnson-kint: Why have these properties been made non-final / mutable?

src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -1292,4 +1292,32 @@ private JsonObject loanRequest(UUID userId, String statusName) {
       .withStatus(statusName)
       .create();
   }
+
+  @Test
+  public void declaredLostPropertyExists()
```
marcjohnson-kint: What is the intent for this test?

If it is to demonstrate that it is possible to set the `declaredLostDate` property, then this can be combined with other tests that demonstrate creation or updating of a loan.

If it is to demonstrate how a client might declare an item lost during the lifecycle of a loan, then I think that could be better represented by a test that demonstrates that process. In that process, I doubt that the item would be declared lost during loan creation.

descriptors/ModuleDescriptor-template.json
mrdpopov: I think that previously, when the version was increased to 6.3 in raml, was not increased to 6.3 in module descriptor. @marc-johnson Do you think that it's a good idea to bump the version up to 6.4 in both files?

ramls/loan.json
mrdpopov: Storing comments is already implemented using the loan history mechanism.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Yes, please, if there was meant to be a 6.3 and it was missed, please move both onto 6.4

(That likely means that nothing binds directly to 6.3 otherwise there would've been environmental build failures)

ramls/loan.json
marcjohnson-kint: Ok, so the intention is to use the `last action comment` for declared lost comments?

src/test/java/org/folio/rest/api/LoansApiTest.java
mrdpopov: The endpoint for declaring lost is in mod-circulation and that logic is tested there. This is just storage and the test demostrates that the date field can be stored and retrieved.

src/test/java/org/folio/rest/support/builders/LoanRequestBuilder.java
mrdpopov: It is so cumbersome to add new fields(the need to update all the constructors, just to add one field?).  I could not find any justifications of bloating the constructor even more.

marcjohnson-kint: The intent was that any instance of the builder is immutable, meaning it can be used as a prototype / template, so can be shared or used as a base line for multiple operations.

Making it mutable means that all instances of the builder based upon the same chain will also get that change.

I don't really follow the argument that this is just a test class. However I'm not inclined to argue if you are intent on making this change.

ramls/loan.json
mrdpopov: As per the clarificaitons from the PO, `the comments appear in that loan action table. ` - that is the loan history. I don't think that we need to store the last comment specifically.

src/test/java/org/folio/rest/api/LoansApiTest.java
marcjohnson-kint: Given the intent is only to demonstrate that the property can be used, and not demonstrate any aspect of the intended use or process (e.g. if folks don't use mod-circulation) then I don't think this needs to be a separate test, it can be included in the existing create / update tests.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

marcjohnson-kint:
Please resolve the failing test due to the formatting of the property.

ramls/loan.json
marcjohnson-kint: Ok, thank you for explaining that. I'll resolve this.

(Personally I think the PO is likely describing the behaviour from a user experience perspective where the loan action table in the UI that happens to be constructed from the loan history)

src/test/java/org/folio/rest/support/builders/LoanRequestBuilder.java
mrdpopov: Done.

src/test/java/org/folio/rest/api/LoansApiTest.java
mrdpopov: Done.

ramls/loan.json
```diff
@@ -93,6 +93,11 @@
...
+      "description": "Date and time loan was declared lost",
```
marcjohnson-kint: I think it could be good to get wording from Emma to describe this. I don't think the loan is declared lost.

```suggestion
      "description": "Date and time the item was declared lost during this loan",
```

mrdpopov: Done.

mrdpopov: Indeed, the loan is not declared lost, it's the item.

PR #227, opened 2019-12-13
CIRC-598: Fix loan.userId fullTextIndex

julianladisch:
Change dictionary from english to simple.

The upgrade script drops the old index with 'english' dictionary.
RMB only (re-)creates the index with the correct 'simple' dictionary if
an index with that name does not exist.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -5,7 +5,7 @@
   "scripts": [
     {
       "run": "before",
-      "fromModuleVersion": "10.0.1",
+      "fromModuleVersion": "10.1.0",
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1,4 +1,14 @@
 {
+  "fullText": {
+    "defaultDictionary": "simple"
+  },
+  "scripts": [
+    {
+      "run": "before",
+      "fromModuleVersion": "10.0.1",
```
marcjohnson-kint: The next release (from mainline) of mod-circulation-storage is `10.1.0`. I think this should match that version.

If we then back port it to a bug fix release, then that might need to change.

PR #228, opened 2019-12-16
CIRCSTORE-177 Add effective Overdue Fine Policy and Lost Item Policy to loan schema

Svitlana-Zmiivska:
**Purpose**
Add effective Overdue Fine Policy and Lost Item Policy to loan schema

**Link**
[CIRCSTORE-177](https://issues.folio.org/browse/CIRCSTORE-177)

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -98,7 +98,7 @@
       "type" : "string",
       "format": "date-time"
     },
-    "effectiveOverdueFinePolicyId": {
+    "overdueFinePolicyId": {
       "description": "ID of overdue fines policy at the time the item is check-in or renewed",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
       "type": "string"
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -98,6 +98,16 @@
       "type" : "string",
       "format": "date-time"
     },
+    "effectiveOverdueFinePolicyId": {
+      "description": "ID of overdue fines policy at the time the item is check-in or renewed",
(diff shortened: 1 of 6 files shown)


ramls/loan.json
```diff
@@ -98,6 +98,14 @@
       "type" : "string",
       "format": "date-time"
     },
+    "effectiveOverdueFinePolicy": {
```
bohdan-suprun: Is it an Id identifier?

Svitlana-Zmiivska: Good point. It should be an identifier

ramls/loan.json
```diff
@@ -98,6 +98,16 @@
       "type" : "string",
       "format": "date-time"
     },
+    "effectiveOverdueFinePolicyId": {
```
marcjohnson-kint: What is the effective overdue fine policy? Why is that different to the lost item policy?

Svitlana-Zmiivska: The effective overdue fine policy is needed to facilitate the quick determination of overdue fines at the time the item is check-in or renewed. 
The lost item policy is needed to facilitate the quick determination  of when the item ages to lost and the associated fees or the associated fees if the patron declares the item lost.

Svitlana-Zmiivska: You can find more details about these fields in the requirements of the story https://issues.folio.org/browse/CIRC-591

marcjohnson-kint: I don't think I'm explaining myself well.

These properties appear to be for remembering the policies in use for that loan. What I'm trying to understand is why one is prefixed with `effective` and the other isn't.

`effective` is typically used for when a property is derived from related properties in a record hierarchy (e.g. items and holdings, which is the only place I'm aware of that it has been used).

Svitlana-Zmiivska: Changed

PR #230, opened 2020-01-25
CIRCSTORE-184: Add several loan indexes.

julianladisch:
BNCF needs them for open loans queries:
https://folio-project.slack.com/archives/C9BBWRCNB/p1579860339261200

"caseSensitive" and "removeAccents" for username are taken from the definition in mod-users:
https://github.com/folio-org/mod-users/blob/v16.0.0/src/main/resources/templates/db_scripts/schema.json#L80

All other indexes are UUIDs or dates that are case insensitive and do not require accent removal.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -12,6 +12,7 @@
   "tables": [
     {
       "tableName": "loan",
+      "fromModuleVersion": "10.1.0",
       "withMetadata": true,
       "withAuditing": true,
       "auditingTableName": "audit_loan",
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -56,6 +56,36 @@
           "tOps": "ADD",
           "caseSensitive": true,
           "removeAccents": false
+        },
+        {
+          "fieldName": "checkinServicePointId",
```
bohdan-suprun: Could you please add `fromModuleVersion`?

julianladisch: Done.

PR #231, opened 2020-01-27
CIRCSTORE-181: Store instance identifiers in request.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-181: Backend: Search for requests by instance ISBN (or other Identifiers).

* Add `item.identifiers` array;
* Add full text indexes for the new property;
* Add missed indexes for request search box. The query is following:
`((requesterId=="900" or requester.barcode="900*" or item.title="900*" or item.barcode="900*" or itemId=="900")) sortby requestDate`, so need indexes for:
  * requesterId
  * requester.barcode
  * item.title
  * item.barcode
  * itemId

### DIFF
--- a/ramls/uuid.json
+++ b/ramls/uuid.json
@@ -1,0 +1,5 @@
+{
+  "description": "A universally unique identifier (UUID), this is a 128-bit number used to identify a record and is shown in hex with dashes, for example 6312d172-f0cf-40f6-b27d-9fa8feaf332f; the UUID version must be from 1-5; see https://dev.folio.org/guides/uuids/",
+  "type": "string",
+  "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$"
+}
--- a/ramls/uuid.json
+++ b/ramls/uuid.json
@@ -1,5 +1,0 @@
-{
-  "description": "A universally unique identifier (UUID), this is a 128-bit number used to identify a record and is shown in hex with dashes, for example 6312d172-f0cf-40f6-b27d-9fa8feaf332f; the UUID version must be from 1-5; see https://dev.folio.org/guides/uuids/",
-  "type": "string",
-  "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$"
-}
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -83,7 +83,6 @@
       "tableName": "request",
       "withMetadata": true,
       "withAuditing": false,
-      "fromModuleVersion": "10.1.0",
       "foreignKeys": [
         {
           "fieldName": "cancellationReasonId",
@@ -115,37 +114,6 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
-        },
-        {
-          "fieldName": "item.identifiers",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
-          "fieldName": "requesterId",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        }
-      ],
-      "fullTextIndex": [
-        {
-          "fieldName": "item.identifiers",
-          "arraySubfield" : "value",
-          "arrayModifiers": ["identifierTypeId"]
-        },
-        {
-          "fieldName": "item.barcode",
-          "tOps": "ADD"
-        },
-        {
-          "fieldName": "item.title",
-          "tOps": "ADD"
-        },
-        {
-          "fieldName": "requester.barcode",
-          "tOps": "ADD"
         }
       ],
       "customSnippetPath": "requestUpdateTrigger.sql"
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -83,6 +83,7 @@
       "tableName": "request",
       "withMetadata": true,
       "withAuditing": false,
+      "fromModuleVersion": "10.1.0",
       "foreignKeys": [
         {
           "fieldName": "cancellationReasonId",
@@ -114,6 +115,37 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "item.identifiers",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
+        },
+        {
+          "fieldName": "requesterId",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
+        }
+      ],
+      "fullTextIndex": [
+        {
+          "fieldName": "item.identifiers",
+          "arraySubfield" : "value",
+          "arrayModifiers": ["identifierTypeId"]
+        },
+        {
+          "fieldName": "item.barcode",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "item.title",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "requester.barcode",
+          "tOps": "ADD"
         }
       ],
       "customSnippetPath": "requestUpdateTrigger.sql"
--- a/src/test/java/org/folio/rest/api/RequestsApiTest.java
+++ b/src/test/java/org/folio/rest/api/RequestsApiTest.java
@@ -101,11 +101,11 @@ public class RequestsApiTest extends ApiTests {
     DateTime holdShelfExpirationDate = new DateTime(2017, 8, 31, 0, 0, DateTimeZone.UTC);
 
     UUID isbnIdentifierId = UUID.randomUUID();
-    UUID bnbIdentifierId = UUID.randomUUID();
+    UUID issnIdentifierId = UUID.randomUUID();
 
     final ItemSummary nod = new ItemSummary("Nod", "565578437802")
       .addIdentifier(isbnIdentifierId, "978-92-8011-566-9")
-      .addIdentifier(bnbIdentifierId, "2193988");
+      .addIdentifier(issnIdentifierId, "2193988");
 
     JsonObject representation = createEntity(
       new RequestRequestBuilder()
@@ -155,7 +155,7 @@ public class RequestsApiTest extends ApiTests {
     assertThat(identifiers.getJsonObject(0).getString("value"),
       is("978-92-8011-566-9"));
     assertThat(identifiers.getJsonObject(1).getString("identifierTypeId"),
-      is(bnbIdentifierId.toString()));
+      is(issnIdentifierId.toString()));
     assertThat(identifiers.getJsonObject(1).getString("value"), is("2193988"));
 
     assertThat(representation.containsKey("requester"), is(true));
--- a/src/test/java/org/folio/rest/api/RequestsApiTest.java
+++ b/src/test/java/org/folio/rest/api/RequestsApiTest.java
@@ -1707,7 +1707,7 @@ public class RequestsApiTest extends ApiTests {
       requestStorageUrl());
 
     List<JsonObject> isbnRequests = findRequestsByQuery(
-      "item.identifiers = /@identifierTypeId=%s %s", isbnIdentifierId, isbn);
+      "item.identifiers = %s and item.identifiers = %s", isbnIdentifierId, isbn);
 
     assertThat(isbnRequests.size(), is(2));
     assertThat(isbnRequests.get(0).getString("id"), is(nodRequestId.toString()));
--- a/src/test/java/org/folio/rest/api/RequestsApiTest.java
+++ b/src/test/java/org/folio/rest/api/RequestsApiTest.java
@@ -19,6 +19,7 @@ import static org.folio.rest.support.builders.RequestRequestBuilder.OPEN_AWAITIN
 import static org.folio.rest.support.builders.RequestRequestBuilder.OPEN_AWAITING_PICKUP;
 import static org.folio.rest.support.builders.RequestRequestBuilder.OPEN_IN_TRANSIT;
 import static org.folio.rest.support.builders.RequestRequestBuilder.OPEN_NOT_YET_FILLED;
+import static org.folio.rest.support.builders.RequestRequestBuilder.ItemSummary;
 import static org.folio.rest.support.matchers.TextDateTimeMatcher.equivalentTo;
 import static org.folio.rest.support.matchers.TextDateTimeMatcher.withinSecondsAfter;
 import static org.folio.rest.support.matchers.ValidationErrorMatchers.hasMessage;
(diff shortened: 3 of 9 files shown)


ramls/uuid.json
```diff
@@ -0,0 +1,5 @@
+{
+  "description": "A universally unique identifier (UUID), this is a 128-bit number used to identify a record and is shown in hex with dashes, for example 6312d172-f0cf-40f6-b27d-9fa8feaf332f; the UUID version must be from 1-5; see https://dev.folio.org/guides/uuids/",
+  "type": "string",
+  "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$"
```
marcjohnson-kint: Please check if there are other usages of this regular expression and replace them with this schema

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -114,6 +115,37 @@
...
+          "fieldName": "item.barcode",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "item.title",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "requester.barcode",
+          "tOps": "ADD"
```
marcjohnson-kint: Why have indexes been added for properties other than the item identifiers?

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -97,6 +100,13 @@ public void canCreateARequest()
     DateTime requestExpirationDate = new DateTime(2017, 7, 30, 0, 0, DateTimeZone.UTC);
     DateTime holdShelfExpirationDate = new DateTime(2017, 8, 31, 0, 0, DateTimeZone.UTC);
 
+    UUID isbnIdentifierId = UUID.randomUUID();
+    UUID bnbIdentifierId = UUID.randomUUID();
```
marcjohnson-kint: What is a BNB identifier?

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -1639,6 +1662,59 @@ public void awaitingPickupRequestClosedDateIsNotPresentAfterStatusUpdateFromOpen
...
+
+    final UUID nodRequestId = UUID.randomUUID();
+    final UUID smallAngryPlanetRequestId = UUID.randomUUID();
+    final UUID issnIdentifierId = UUID.randomUUID();
+    final UUID isbnIdentifierId = UUID.randomUUID();
+    final String isbn = "978-92-8011-566-10";
+
+    final ItemSummary nod = new ItemSummary("Nod", "565578437802")
+      .addIdentifier(issnIdentifierId, "978-92-8011-566-9")
+      .addIdentifier(UUID.randomUUID(), "2193988")
```
marcjohnson-kint: Why are the random UUIDs inline on this line and a constant above?

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
```diff
@@ -133,6 +141,14 @@ public JsonObject create() {
       put(item, "title", this.itemSummary.title);
       put(item, "barcode", this.itemSummary.barcode);
 
+      final JsonArray identifiers = new JsonArray(this.itemSummary.identifiers
```
marcjohnson-kint: What happens if itemSummary.identifiers is null?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: These properties are used in request search box, actually the search query is following: 
`((requesterId=="900" or requester.barcode="900*" or item.title="900*" or item.barcode="900*" or itemId=="900")) sortby requestDate`

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
```diff
@@ -783,13 +803,30 @@ public RequestRequestBuilder withTags(Tags tags) {
       tags);
   }
 
-  private class ItemSummary {
+  public static class ItemSummary {
```
marcjohnson-kint: Why is this class static, that suggests there can only ever be a single instance of this class?

src/test/java/org/folio/rest/api/RequestsApiTest.java
bohdan-suprun: I've found it on a sample instance, I'm not sure what this means 😄.

marcjohnson-kint: I suggest using a more common identifier type for this example, like ISSN

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
bohdan-suprun: It should not be null. It is initialized by an empty array and there is no setter for it (there is only an `addIdentifier` method).

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Hmm, I'm not sure we should add indexes in unrelated pull requests only because they are used in the search.

@skoczko @julianladisch @adamdickmeiss What do you think, should we back fill indexes for these properties?

src/test/java/org/folio/rest/api/RequestsApiTest.java
mreno-EBSCO: Looks like it is the British National Bibliography ID: https://www.bl.uk/collection-metadata/metadata-services

I learned something new. Good thing I still get these PR emails. 😄

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
bohdan-suprun: Because it is now possible to instantiate it in a test method and pass as a parameter to `RequestRequestBuilder#withItem` method.

src/test/java/org/folio/rest/api/RequestsApiTest.java
bohdan-suprun: It is actually means 'an another identifier, in which we're not interesting in'.

src/test/java/org/folio/rest/api/RequestsApiTest.java
marcjohnson-kint: @mreno-EBSCO  It is probably time to review the circulation code owners. If you or @mtraneis want to leave the group, please let me know (I'm happy for you both to remain)

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
marcjohnson-kint: Does it need to be static to do that? Is that how nested classes work when they are made public?

bohdan-suprun: Non-static nested class can not be instantiated without an instance of outer class. Commonly, non-static classes used when they have to access to outer class's properties/methods/etc. An example can be `Itr` implementation in `ArrayList` when the `Itr` have to access to array of elements of the list. It also requires a bit complicated constructions: `new RequestRequestBuilder.new ItemSummary`.

Static nested classes does not use internal state of outer class, it is more like an additional namespace, and makes thinks a bit easier.

ramls/uuid.json
bohdan-suprun: I'll check request.json schema only. Or do you mean all the json schemes?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: You're right. Should I fill in a bug to create indexes?

src/test/java/org/folio/rest/support/builders/RequestRequestBuilder.java
marcjohnson-kint: Thanks for the explanation. I think I might have moved it to a separate class.

src/test/java/org/folio/rest/api/RequestsApiTest.java
marcjohnson-kint: Maybe extract a constant to describe that intent.

I thought the test did search for that identifier?

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: How do we know that it is a bug?

ramls/uuid.json
marcjohnson-kint: I think it's worth checking all of them. This is effectively like extracting a method to remove duplication.

bohdan-suprun: Hi @marcjohnson-kint,

In the uuid.json I'm using pattern described here https://dev.folio.org/guides/uuids/.

But all existing UUIDs uses weaker pattern:
`^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$` - used now
`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$` - used in `uuid.json`

I don't think that we can just replace all UUID patterns with the new one, it should be considered as a breaking change.

Should I maybe use the weak UUID pattern in the `uuid.json` file?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Maybe not, however RMB logs a warning that full text search are being performed without index.

src/test/java/org/folio/rest/api/RequestsApiTest.java
bohdan-suprun: Ok, sure. No, the test do a search for ISBN identifier.

ramls/uuid.json
marcjohnson-kint: > Hi @marcjohnson-kint,
> 
> In the uuid.json I'm using pattern described here https://dev.folio.org/guides/uuids/.

That is correct.

> But all existing UUIDs uses weaker pattern:
> `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$` - used now
> `^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$` - used in `uuid.json`
> 
> I don't think that we can just replace all UUID patterns with the new one, it should be considered as a breaking change.

Agreed, this would be breaking compatibility. we shouldn't change any weak UUID regular expressions during this change.

> Should I maybe use the weak UUID pattern in the `uuid.json` file?

I don't think we should use the weak UUID pattern for the shared file.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: @skoczko @julianladisch @adamdickmeiss Is this considered a bug?

marcjohnson-kint: @bohdan I think it is worth asking @julianladisch about the upcoming normalisation changes for identifiers in inventory, if those require changes to indexes, then it might not be worth adding any here only to have to change them soon.

bohdan-suprun: @julianladisch responded that the normalisation work will require index re-create (see RMB-499).

Because of this, I've removed all the previously added indexes. Please note, that CQL search query may be different with/without index, depending on which approach we will chose on UI for search:
* If we allow user to limit search for particular identifier type (e.g. search by ISBN only, as we do in inventory) - then `/@` is not supported without index. With index the query will be `item.identifiers = /@identifierTypeId = 'id' 'value'` but without index this feature is not supported and have to be something like `item.identifiers = 'id' and item.identifiers = 'value'`
* In case the search will not require limitation by identifier type then the CQL will be the same.

src/test/java/org/folio/rest/support/builders/RequestItemSummary.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public RequestItemSummary addIdentifier(UUID identifierId, String value) {
+    final List<Pair<UUID, String>> copiedIdentifiers = new ArrayList<>(identifiers);
+    copiedIdentifiers.add(new ImmutablePair<>(identifierId, value));
+
+    return new RequestItemSummary(
+      this.title,
+      this.barcode,
+      copiedIdentifiers
+    );
```
marcjohnson-kint: Personally, I would close these brackets on the line above

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Hmm, maybe I've approved this too soon.

@bohdan-suprun Please can you follow up with @julianladisch about how the query will work in inventory when the value has been normalised.

Ideally, the client wouldn't have to change it's query between these changes.

If they will have to, we might want to hold off merging this and apply the normalization within the same pull request.

bohdan-suprun: I don't think that the normalization will affect CQL query. My main point that the CQL query will be different when we use an index (either normalized or not) or don't.

This depends on which search approach we will chose on UI:
* If we chose the approach when we have a drop-down to pick an identifier type, like what we currently have in inventory:
image
then the CQL will be different with and without index.
* If the search box is a simple keyword-like search (what we have currently in requests, when we search is `requesterId = '...' or item.barcode = '...' or etc...`) then CQL remains the same for both cases.

marcjohnson-kint: @bohdan-suprun My understanding is that the normalization will work by interpreting some CQL queries differently.

I think that means we need to understand what kind of queries will need to be used after the normalization changes are made.

The query (or queries) and indexes should match those in inventory, after the normalization changes.

Does that make sense?

bohdan-suprun: I've posted a question in RMB-499.

julianladisch: My suggestion is to merge this pull request and not to wait for RMB-499.
RMB-499 will be released in RMB v29.3.x.
mod-circulation-storage uses RMB v26.2.3 and requires an upgrade: https://issues.folio.org/browse/CIRCSTORE-183
Using the barcode normalization from RMB-499 requires a new index. And it probably requires changes to CQL to use the (virtual) normalized field.

ramls/uuid.json
```diff
@@ -0,0 +1,5 @@
...
+}
```
julianladisch: Consider upgrading raml-util submodule and using
https://github.com/folio-org/raml/blob/raml1.0/schemas/uuid.schema

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Agreed, this should be merged to allow for the work to populate these properties to be done.

A new issue needs to be raised for supporting the appropriate queries when the normalisation work is ready. @bohdan-suprun please can you raise that?

bohdan-suprun:
> There is still a very minor outstanding comment from me and a comment from Julian about updating the RAML definitions.

Thanks, it is fixed now.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: The https://issues.folio.org/browse/CIRCSTORE-186 has been created.

PR #232, opened 2020-01-31
CIRCSTORE-183: Upgrade RMB from 26.2.3 to 29.2.x

kaladay:
Updating to the latest RMB required significant changes.

Summary:
- Use HttpStatus more consistently.
- Perform update requirements as described in RMB upgrade cookbook.
- Remove/update/replace deprecated code/functionality.
- Address numerous issues with tests, adding new code to address behavioral changes.
- Attempt a less fragile error handling design.
- Ensure the `@validate` is in use by LoanPoliciesAPI.java.
- Rewrite tests to address change in behavior as defined in RMB-459 and RMB-353.
- Fixed random typos in the code that I observed in files that I am already changing (such as `udpatePreference`).

Of particular note, there was some discussion regarding the handling of error messages.
The discussed approach was to encourage the use of RMB's error messages.
However, after extensive review it seems that there are sometimes inconsistent status codes (500 that should be 422 or 400 that should be a 422) and multiple different messages for the same error situation.
I strongly suggest specifically reviewing the approach provided here.

see: https://issues.folio.org/browse/CIRCSTORE-183
see: https://issues.folio.org/browse/UIREQ-401
see: https://issues.folio.org/browse/RMB-459
see: https://issues.folio.org/browse/RMB-353

### DIFF
--- a/ramls/loan-policy-storage.raml
+++ b/ramls/loan-policy-storage.raml
@@ -11,11 +11,13 @@ documentation:
 types:
   loan-policy: !include loan-policy.json
   loan-policies: !include loan-policies.json
+  errors: !include raml-util/schemas/errors.schema
 
 traits:
   language: !include raml-util/traits/language.raml
   pageable: !include raml-util/traits/pageable.raml
   searchable: !include raml-util/traits/searchable.raml
+  validate: !include raml-util/traits/validation.raml
 
 resourceTypes:
   collection: !include raml-util/rtypes/collection.raml
@@ -39,6 +41,7 @@ resourceTypes:
         501:
           description: "Not implemented yet"
     post:
+      is: [validate]
       responses:
         501:
           description: "Not implemented yet"
--- a/src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
@@ -5,13 +5,17 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import org.folio.rest.annotations.Validate;
+import org.folio.rest.jaxrs.model.Error;
+import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.LoanPolicies;
 import org.folio.rest.jaxrs.model.LoanPolicy;
 import org.folio.rest.jaxrs.resource.LoanPolicyStorage;
 import org.folio.rest.persist.MyPgUtil;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.TenantTool;
+import org.folio.rest.tools.utils.ValidationHelper;
 import javax.ws.rs.core.Response;
 import java.util.Map;
 import static org.folio.rest.impl.Headers.TENANT_HEADER;
@@ -48,6 +52,7 @@ public class LoanPoliciesAPI implements LoanPolicyStorage {
   }
 
   @Override
+  @Validate
   public void getLoanPolicyStorageLoanPolicies(
     int offset,
     int limit,
@@ -71,7 +76,16 @@ public class LoanPoliciesAPI implements LoanPolicyStorage {
     Context vertxContext) {
 
     PgUtil.post(LOAN_POLICY_TABLE, entity, okapiHeaders, vertxContext,
-        PostLoanPolicyStorageLoanPoliciesResponse.class, asyncResultHandler);
+        PostLoanPolicyStorageLoanPoliciesResponse.class, reply -> {
+      if (isInvalidUUIDError(reply)) {
+        asyncResultHandler.handle(
+          Future.succeededFuture(LoanPolicyStorage.PostLoanPolicyStorageLoanPoliciesResponse
+            .respond422WithApplicationJson(invalidUUIDError(entity))));
+        return;
+      }
+
+      asyncResultHandler.handle(reply);
+    });
   }
 
   @Override
@@ -111,5 +125,38 @@ public class LoanPoliciesAPI implements LoanPolicyStorage {
 
     MyPgUtil.putUpsert204(LOAN_POLICY_TABLE, entity, loanPolicyId, okapiHeaders, vertxContext,
         PutLoanPolicyStorageLoanPoliciesByLoanPolicyIdResponse.class, asyncResultHandler);
+  }
+
+  private Errors invalidUUIDError(LoanPolicy entity) {
+    return ValidationHelper.createValidationErrorMessage(
+      "itemId", entity.getId(),
+      "Provided UUID for the lone policy is invalid");
+  }
+
+  private boolean isInvalidUUIDError(AsyncResult<Response> reply) {
+    String message = "";
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
+      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
+
+          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
+            message = error.getMessage().toLowerCase();
+        }
+      } else {
+        message = reply.result().getEntity().toString().toLowerCase();
+      }
+    }
+
+    return message.contains("invalid input syntax for type uuid: ") ||
+      message.contains("duplicate key value violates unique constraint ");
   }
 }
--- a/src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
@@ -5,17 +5,13 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import org.folio.rest.annotations.Validate;
-import org.folio.rest.jaxrs.model.Error;
-import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.LoanPolicies;
 import org.folio.rest.jaxrs.model.LoanPolicy;
 import org.folio.rest.jaxrs.resource.LoanPolicyStorage;
 import org.folio.rest.persist.MyPgUtil;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
-import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.TenantTool;
-import org.folio.rest.tools.utils.ValidationHelper;
 import javax.ws.rs.core.Response;
 import java.util.Map;
 import static org.folio.rest.impl.Headers.TENANT_HEADER;
@@ -76,16 +72,7 @@ public class LoanPoliciesAPI implements LoanPolicyStorage {
     Context vertxContext) {
 
     PgUtil.post(LOAN_POLICY_TABLE, entity, okapiHeaders, vertxContext,
-        PostLoanPolicyStorageLoanPoliciesResponse.class, reply -> {
-      if (isInvalidUUIDError(reply)) {
-        asyncResultHandler.handle(
-          Future.succeededFuture(LoanPolicyStorage.PostLoanPolicyStorageLoanPoliciesResponse
-            .respond422WithApplicationJson(invalidUUIDError(entity))));
-        return;
-      }
-
-      asyncResultHandler.handle(reply);
-    });
+        PostLoanPolicyStorageLoanPoliciesResponse.class, asyncResultHandler);
   }
 
   @Override
@@ -125,38 +112,5 @@ public class LoanPoliciesAPI implements LoanPolicyStorage {
 
     MyPgUtil.putUpsert204(LOAN_POLICY_TABLE, entity, loanPolicyId, okapiHeaders, vertxContext,
         PutLoanPolicyStorageLoanPoliciesByLoanPolicyIdResponse.class, asyncResultHandler);
-  }
-
-  private Errors invalidUUIDError(LoanPolicy entity) {
-    return ValidationHelper.createValidationErrorMessage(
-      "itemId", entity.getId(),
-      "Provided UUID for the lone policy is invalid");
-  }
-
-  private boolean isInvalidUUIDError(AsyncResult<Response> reply) {
-    String message = "";
-
-    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
-      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
-
-      // When entity is an instance of Errors, the getEntity().toString()
-      // will return an object identifier. Process the object to correctly
-      // parse the message.
-      if (reply.result().getEntity() instanceof Errors) {
-        Errors errors = (Errors) reply.result().getEntity();
-
-        for (int i = 0; i < errors.getErrors().size(); i++) {
-          Error error = errors.getErrors().get(i);
-
-          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
-            message = error.getMessage().toLowerCase();
-        }
-      } else {
-        message = reply.result().getEntity().toString().toLowerCase();
-      }
-    }
-
-    return message.contains("invalid input syntax for type uuid: ") ||
-      message.contains("duplicate key value violates unique constraint ");
   }
 }
--- a/src/main/java/org/folio/rest/impl/LoansAPI.java
+++ b/src/main/java/org/folio/rest/impl/LoansAPI.java
@@ -29,6 +29,7 @@ import org.folio.rest.jaxrs.resource.LoanStorage;
 import org.folio.rest.persist.MyPgUtil;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.TenantTool;
 import org.folio.rest.tools.utils.ValidationHelper;
 import org.folio.support.ResultHandlerFactory;
@@ -293,9 +294,30 @@ public class LoansAPI implements LoanStorage {
   }
 
   private boolean isMultipleOpenLoanError(AsyncResult<Response> reply) {
-    return reply.succeeded()
-      && reply.result().getStatus() == 400
-      && reply.result().getEntity().toString().contains("loan_itemid_idx_unique");
+    String message = "";
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
+      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
+
+          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
+            message = error.getMessage().toLowerCase();
+        }
+      } else {
+        message = reply.result().getEntity().toString().toLowerCase();
+      }
+    }
+
+    return message.contains("value already exists in table loan: ")
+      || message.contains("duplicate key value violates unique constraint ");
   }
 
   private boolean isOpenAndHasNoUserId(Loan loan) {
@@ -318,8 +340,6 @@ public class LoansAPI implements LoanStorage {
     asyncResultHandler.handle(succeededFuture(
       responseCreator.apply(errors)));
   }
-
-
 
   private String createAnonymizationSQL(
     @NotNull String userId,
--- a/src/main/java/org/folio/rest/impl/LoansAPI.java
+++ b/src/main/java/org/folio/rest/impl/LoansAPI.java
@@ -1,6 +1,7 @@
 package org.folio.rest.impl;
 
 import static io.vertx.core.Future.succeededFuture;
+import static org.folio.HttpStatus.HTTP_BAD_REQUEST;
 import static org.folio.rest.impl.Headers.TENANT_HEADER;
 import static org.folio.support.ModuleConstants.LOAN_CLASS;
 import static org.folio.support.ModuleConstants.LOAN_HISTORY_TABLE;
@@ -18,6 +19,7 @@ import javax.ws.rs.core.Response;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.folio.rest.annotations.Validate;
+import org.folio.rest.impl.util.OkapiResponseUtil;
 import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.Loan;
@@ -29,7 +31,6 @@ import org.folio.rest.jaxrs.resource.LoanStorage;
 import org.folio.rest.persist.MyPgUtil;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
-import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.TenantTool;
 import org.folio.rest.tools.utils.ValidationHelper;
 import org.folio.support.ResultHandlerFactory;
@@ -233,7 +234,7 @@ public class LoansAPI implements LoanStorage {
 
     MyPgUtil.putUpsert204(LOAN_TABLE, loan, loanId, okapiHeaders, vertxContext,
       PutLoanStorageLoansByLoanIdResponse.class, reply -> {
-        if (isMultipleOpenLoanError(reply)) {
+        if (isMultipleOpenLoanErrorOnUpsert(reply)) {
           asyncResultHandler.handle(
             succeededFuture(
               LoanStorage.PutLoanStorageLoansByLoanIdResponse
@@ -293,31 +294,17 @@ public class LoansAPI implements LoanStorage {
       "Cannot have more than one open loan for the same item");
   }
 
+  // Remove/Replace this function when MyPgUtil.putUpsert204() is removed/replaced.
+  private boolean isMultipleOpenLoanErrorOnUpsert(AsyncResult<Response> reply) {
+    return reply.succeeded()
+      && reply.result().getStatus() == HTTP_BAD_REQUEST.toInt()
+      && reply.result().hasEntity()
+      && reply.result().getEntity().toString().contains("loan_itemid_idx_unique");
+  }
+
   private boolean isMultipleOpenLoanError(AsyncResult<Response> reply) {
-    String message = "";
-
-    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
-      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
-
-      // When entity is an instance of Errors, the getEntity().toString()
-      // will return an object identifier. Process the object to correctly
-      // parse the message.
-      if (reply.result().getEntity() instanceof Errors) {
-        Errors errors = (Errors) reply.result().getEntity();
-
-        for (int i = 0; i < errors.getErrors().size(); i++) {
-          Error error = errors.getErrors().get(i);
-
-          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
-            message = error.getMessage().toLowerCase();
-        }
-      } else {
-        message = reply.result().getEntity().toString().toLowerCase();
-      }
-    }
-
-    return message.contains("value already exists in table loan: ")
-      || message.contains("duplicate key value violates unique constraint ");
+    return OkapiResponseUtil.containsErrorMessage(
+      reply, "value already exists in table loan: ");
   }
 
   private boolean isOpenAndHasNoUserId(Loan loan) {
--- a/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
+++ b/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
@@ -8,11 +8,13 @@ import javax.ws.rs.core.Response;
 
 import org.folio.HttpStatus;
 import org.folio.rest.annotations.Validate;
+import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.RequestPreference;
 import org.folio.rest.jaxrs.model.RequestPreferences;
 import org.folio.rest.jaxrs.resource.RequestPreferenceStorage;
 import org.folio.rest.persist.PgUtil;
+import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.ValidationHelper;
 
 import io.vertx.core.AsyncResult;
@@ -66,9 +68,30 @@ public class RequestPreferencesAPI implements RequestPreferenceStorage {
   }
 
   private boolean isUniqueUserIdViolation(AsyncResult<Response> reply) {
-    return reply.succeeded() &&
-      reply.result().getStatus() == HttpStatus.HTTP_BAD_REQUEST.toInt() &&
-      reply.result().getEntity().toString().contains("user_request_preference_userid_idx_unique");
+    String message = "";
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
+      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
+
+          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
+            message = error.getMessage().toLowerCase();
+        }
+      } else {
+        message = reply.result().getEntity().toString().toLowerCase();
+      }
+    }
+
+    return message.contains(" value already exists in table ") ||
+      message.contains("user_request_preference_userid_idx_unique");
   }
 
   private Errors preferenceAlreadyExistsError(String userId) {
--- a/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
+++ b/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
@@ -1,20 +1,19 @@
 package org.folio.rest.impl;
 
+import static org.folio.HttpStatus.HTTP_UNPROCESSABLE_ENTITY;
 import static io.vertx.core.Future.succeededFuture;
 
 import java.util.Map;
 
 import javax.ws.rs.core.Response;
 
-import org.folio.HttpStatus;
 import org.folio.rest.annotations.Validate;
-import org.folio.rest.jaxrs.model.Error;
+import org.folio.rest.impl.util.OkapiResponseUtil;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.RequestPreference;
 import org.folio.rest.jaxrs.model.RequestPreferences;
 import org.folio.rest.jaxrs.resource.RequestPreferenceStorage;
 import org.folio.rest.persist.PgUtil;
-import org.folio.rest.tools.RTFConsts;
 import org.folio.rest.tools.utils.ValidationHelper;
 
 import io.vertx.core.AsyncResult;
@@ -59,7 +58,8 @@ public class RequestPreferencesAPI implements RequestPreferenceStorage {
     return reply -> {
       if (isUniqueUserIdViolation(reply)) {
         asyncResultHandler.handle(
-          succeededFuture(Response.status(HttpStatus.HTTP_UNPROCESSABLE_ENTITY.toInt()).header("Content-Type", "application/json")
+          succeededFuture(Response.status(HTTP_UNPROCESSABLE_ENTITY.toInt())
+            .header("Content-Type", "application/json")
             .entity(preferenceAlreadyExistsError(entity.getUserId())).build()));
       } else {
         asyncResultHandler.handle(reply);
@@ -68,30 +68,8 @@ public class RequestPreferencesAPI implements RequestPreferenceStorage {
   }
 
   private boolean isUniqueUserIdViolation(AsyncResult<Response> reply) {
-    String message = "";
-
-    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
-      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
-
-      // When entity is an instance of Errors, the getEntity().toString()
-      // will return an object identifier. Process the object to correctly
-      // parse the message.
-      if (reply.result().getEntity() instanceof Errors) {
-        Errors errors = (Errors) reply.result().getEntity();
-
-        for (int i = 0; i < errors.getErrors().size(); i++) {
-          Error error = errors.getErrors().get(i);
-
-          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
-            message = error.getMessage().toLowerCase();
-        }
-      } else {
-        message = reply.result().getEntity().toString().toLowerCase();
-      }
-    }
-
-    return message.contains(" value already exists in table ") ||
-      message.contains("user_request_preference_userid_idx_unique");
+    return OkapiResponseUtil.containsErrorMessage(
+      reply, " value already exists in table ");
   }
 
   private Errors preferenceAlreadyExistsError(String userId) {
--- a/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
+++ b/src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
@@ -6,6 +6,7 @@ import java.util.Map;
 
 import javax.ws.rs.core.Response;
 
+import org.folio.HttpStatus;
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.Errors;
 import org.folio.rest.jaxrs.model.RequestPreference;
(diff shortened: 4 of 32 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -31,33 +28,13 @@
         }
       ],
       "ginIndex": [
-        {
```
bohdan-suprun: Is it removed because new RMB version provides it out of the box?

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -71,7 +76,16 @@ public void postLoanPolicyStorageLoanPolicies(
     Context vertxContext) {
 
     PgUtil.post(LOAN_POLICY_TABLE, entity, okapiHeaders, vertxContext,
-        PostLoanPolicyStorageLoanPoliciesResponse.class, asyncResultHandler);
+        PostLoanPolicyStorageLoanPoliciesResponse.class, reply -> {
+      if (isInvalidUUIDError(reply)) {
```
bohdan-suprun: We should add UUID pattern to the `loan-policy.json` instead:
`"$ref": "raml-util/schemas/uuid.schema"`

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -293,9 +294,30 @@ private Errors moreThanOneOpenLoanError(Loan entity) {
   }
 
   private boolean isMultipleOpenLoanError(AsyncResult<Response> reply) {
-    return reply.succeeded()
-      && reply.result().getStatus() == 400
-      && reply.result().getEntity().toString().contains("loan_itemid_idx_unique");
+    String message = "";
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
```
bohdan-suprun: I think we should extract it to a separate utility class.

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -293,9 +294,30 @@ private Errors moreThanOneOpenLoanError(Loan entity) {
...
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
+
+          if (error.getType() == RTFConsts.VALIDATION_FIELD_ERROR)
```
bohdan-suprun: Maybe not rely on a specific error type. I don't think that we will have multiply errors here.

src/test/java/org/folio/rest/api/FixedDueDateApiTest.java
```diff
@@ -231,7 +232,7 @@ public void canCreateSchedule()
       ResponseHandler.json(updateCompleted4));
     JsonResponse updateCompleted4Response = updateCompleted4.get(5, TimeUnit.SECONDS);
     assertThat(String.format("Failed to create due date: %s", fixDueDate5.encodePrettily()),
-      updateCompleted4Response.getStatusCode(), is(HttpURLConnection.HTTP_BAD_REQUEST));*/
+      updateCompleted4Response.getStatusCode(), is(HttpStatus.HTTP_UNPROCESSABLE_ENTITY.toInt()));*/
```
bohdan-suprun: Maybe remove it at all, since it is commented.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
bohdan-suprun: Hmm... it is going to be a breaking change then... Is this logic has introduced to fix `org.folio.rest.api.LoanPoliciesApiTest#loanPolicyWithFixedDate`?

ramls/loan-policy-storage.raml
```diff
@@ -39,6 +41,7 @@ resourceTypes:
         501:
           description: "Not implemented yet"
     post:
+      is: [validate]
```
marcjohnson-kint: This changes the API definition, which needs to be reflected in a version change. I'm not sure whether this should be considered a breaking compatible change or not.

src/main/java/org/folio/rest/impl/LoansAPI.java
```diff
@@ -293,9 +294,30 @@ private Errors moreThanOneOpenLoanError(Loan entity) {
...
+            message = error.getMessage().toLowerCase();
+        }
+      } else {
+        message = reply.result().getEntity().toString().toLowerCase();
+      }
+    }
+
+    return message.contains("value already exists in table loan: ")
+      || message.contains("duplicate key value violates unique constraint ");
```
marcjohnson-kint: My interpretation of this change is that the description of the error is more stable than the name of the index. Is that a reasonable understanding?

What happens if the loan table has more than one unique constraint, will this code think that it is multiple open loans?

src/main/java/org/folio/rest/impl/RequestPreferencesAPI.java
```diff
@@ -56,7 +59,7 @@ public void putRequestPreferenceStorageRequestPreferenceById(String id, String l
     return reply -> {
       if (isUniqueUserIdViolation(reply)) {
         asyncResultHandler.handle(
-          succeededFuture(Response.status(422).header("Content-Type", "application/json")
+          succeededFuture(Response.status(HttpStatus.HTTP_UNPROCESSABLE_ENTITY.toInt()).header("Content-Type", "application/json")
```
marcjohnson-kint: Personally, I might statically import this constant to shorten this line. 

Please can it be split to conform with the maximum line length.

```suggestion
          succeededFuture(Response.status(HTTP_UNPROCESSABLE_ENTITY.toInt()).header("Content-Type", "application/json")
```

src/main/java/org/folio/rest/impl/RequestsAPI.java
```diff
@@ -134,9 +136,28 @@ private Errors samePositionInQueueError(Request request) {
...
 
   private boolean isSamePositionInQueueError(AsyncResult<Response> reply) {
-    return reply.succeeded()
-      && reply.result().getStatus() == 400
-      && RequestsApiUtil
-      .hasSamePositionConstraintViolated(reply.result().getEntity().toString());
+    String message = null;
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400 &&
+      reply.result().getStatus() < 600 && reply.result().hasEntity()) {
```
marcjohnson-kint: Why is this a check between 400 and 600, are client and server errors treated the same by PgUtil?

src/test/java/org/folio/rest/api/CancellationReasonsApiTest.java
```diff
@@ -256,7 +258,7 @@ public void canDeleteCancellationRequest()
     assertCreateCancellationReason(request);
     assertDeleteCancellationReason(id);
     JsonResponse getResponse = getCancellationReason(id);
-    assertTrue(getResponse.getStatusCode() == 404);
+    assertEquals(HttpStatus.HTTP_NOT_FOUND.toInt(), getResponse.getStatusCode());
```
marcjohnson-kint: I'm starting to think that a custom matcher that expects a FOLIO `HttpStatus` and the actual value is either the status code or the response itself, might be useful in improving the readability of these assertions.

src/test/java/org/folio/rest/api/FixedDueDateApiTest.java
marcjohnson-kint: Yeah, I think we should remove this commented out code

src/test/java/org/folio/rest/api/FixedDueDateApiTest.java
```diff
@@ -91,7 +92,7 @@ public void canCreateSchedule()
       ResponseHandler.empty(updateBad3Completed));
     Response updateBad3Response = updateBad3Completed.get(5, TimeUnit.SECONDS);
     assertThat(String.format("Failed to create due date: %s", representation.encodePrettily()),
-      updateBad3Response.getStatusCode(), is(422));
+      updateBad3Response.getStatusCode(), is(HttpStatus.HTTP_UNPROCESSABLE_ENTITY.toInt()));
```
marcjohnson-kint: I might be tempted to statically import these constants to improve the readability of these assertions

src/test/java/org/folio/rest/support/WebClient.java
```diff
@@ -5,30 +5,32 @@
...
-import io.vertx.core.http.HttpClientResponse;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.Json;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.web.client.HttpRequest;
+import io.vertx.ext.web.client.HttpResponse;
 
-public class HttpClient {
+public class WebClient {
```
marcjohnson-kint: Why has this class been renamed?

src/main/resources/templates/db_scripts/schema.json
kaladay: Unless I am mis-reading this, according to https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-271 

> Remove each foreign key field index and each primary key field id index...

src/main/java/org/folio/rest/impl/LoansAPI.java
kaladay: I was wondering the same and was uncertain so I left it to review. This will be done.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Ok, thanks for providing the doc.

I think the `userId` and `itemId` have not to be removed. The item and user tables are actually in another DB schemes (diku_mod_inventory_storage and diku_mod_users) so we don't have FKs for them.

Note: usually FKs are declared in schema.json under `tables.foreignKeys` (for example: https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/schema.json#L318).

src/main/java/org/folio/rest/impl/RequestsAPI.java
kaladay: This is just an attempt to make it less fragile. That said, I am fine with making it very specific to exact error codes.

The problem seems to be that sometimes we get 400, 422, and 500 of which, apparently, was all returning the same codes prior to the update based for the original tests.

There is also a situation where the deprecated MyPgUtil upsert is used and that operates with different return objects, codes, and messages than PgUtil.

src/test/java/org/folio/rest/support/WebClient.java
kaladay: According to https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-271 , HttpClient is deprecated in with WebClient.

This test utility class was simply helping provide HttpClient.
Now that it is providing WebClient, it makes more sense to me to update the name to communicate to the coder what this library is now wrapping and using.

src/main/resources/templates/db_scripts/schema.json
kaladay: Ok, I was reading it as `(foreign key field index) and (each primary key field id index)`, which from what I gather should instead be read as `foreign key (field index and each primary key field id index)`? (note the parenthesis inteded to communicate logical grouping.)

Those ginIndex are indexes for columns that act as a primary key.
If I am understanding correctly, then only the foreign key group associated with an index/pkey should be removed?

bohdan-suprun: I think it means following:
* remove all the indexes for primary key (it is index for `id`, I think) - the reason of this I think that RMB now automatically creates such index. 
* remove indexes for properties that have declared in `foreignKeys` section - again, it seems RMB now creates indexes for them automatically.

I don't think that we have to update properties that act as a FK but don't have associated FK constraint.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
kaladay: The logic was indeed added to fix the tests in the same way that kind of logic was already being used elswhere (such as `isSamePositionInQueueError()` from `RequestsAPI.java`).

bohdan-suprun: Ok, I think we can do the following then:
* Fix `loanPolicyWithFixedDate` test by providing a random UUID instead of the invalid `1234567890` string;
* Raise a separate jira to add UUID pattern for the `loan.fixedDueDateScheduleId` property (it will be considered as a breaking change).

@marcjohnson-kint what do you think?

src/main/java/org/folio/rest/impl/LoansAPI.java
kaladay: Yes/No, the index is no longer appearing in the error message in some circumstances.

This messsage itself is (sometimes) also being added: https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L332 .

It could be that this should be now using `table:value` syntax instead of index in addition to the generic error message string `value already exists in table loan:`.
The problem I see then is that we need to change how that error is caught in the first place.
If I remember correctly, there are different tables/values/indexes that would get called by say `isMultipleOpenLoanError`.

It might indeed make more sense to write this as a utility (as requested) and then require a seperate parameter for the specific _table_ and the specific _value_ to be matched.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
marcjohnson-kint: Hmm, am I misunderstanding this part of the change.

An invalid UUID should be different to a unique constraint error, are they being reported as the same?

src/main/java/org/folio/rest/impl/LoansAPI.java
marcjohnson-kint: > Yes/No, the index is no longer appearing in the error message in some circumstances.
> 
> This messsage itself is (sometimes) also being added: https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L332 .

Hmm, it is interesting that the error message is different even under the same circumstances. Can it be confirmed that the underlying cause is the same unique constraint in all cases?

> It could be that this should be now using `table:value` syntax instead of index in addition to the generic error message string `value already exists in table loan:`.
> The problem I see then is that we need to change how that error is caught in the first place.

Can you expand upon what you mean here? Do you mean the error should be reported differently from PgUtil?

> If I remember correctly, there are different tables/values/indexes that would get called by say `isMultipleOpenLoanError`.

The intent of `isMultipleOpenLoanError` was to infer from an opaque error message whether that was the underlying cause of the error.

I think we still need a way to achieve that. I don't know what information we have available in order to make that decision.

> It might indeed make more sense to write this as a utility (as requested) and then require a seperate parameter for the specific _table_ and the specific _value_ to be matched.

If we can extract the general parts of this into a common function with some parameters that would be great.

Some work to do that was done in a special matcher for tests, which might be worth finding. It wouldn't be appropriate to use, as it is test specific.

src/main/java/org/folio/rest/impl/RequestsAPI.java
marcjohnson-kint: > This is just an attempt to make it less fragile. That said, I am fine with making it very specific to exact error codes.
> 
> The problem seems to be that sometimes we get 400, 422, and 500 of which, apparently, was all returning the same codes prior to the update based for the original tests.

I don't think we need to be too specific. I think we do want to separate 4xx and 5xx errors.

Does the same underlying cause (e.g. multiple open loans or request in the same position) sometimes cause PgUtil to report a 4xx and 5xx error?

> There is also a situation where the deprecated MyPgUtil upsert is used and that operates with different return objects, codes, and messages than PgUtil.

Can the `MyPgUtil` use be replaced with `PgUtil` within this pull request?

src/test/java/org/folio/rest/support/WebClient.java
marcjohnson-kint: > According to https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-271 , HttpClient is deprecated in with WebClient.

I think that guidance refers to changes in vert.x, where some methods from HttpClient were deprecated. Maybe RAML Module Builder is advocating migration from vert.x HttpClient to WebClient?

> This test utility class was simply helping provide HttpClient.
> Now that it is providing WebClient, it makes more sense to me to update the name to communicate to the coder what this library is now wrapping and using.

I think I interpret the intent of this class differently. Rather than intended to provide the underlying vert.x HttpClient, I thought it was trying to hide it away from the rest of the test code.

Personally, I'd prefer that the client test code was ignorant as to whether the internal implementation is using vert.x HttpClient, WebClient or a different library.

Could we come up with a name that better describes the intent of this wrapper? In mod-circulation, it is called OkapiHttpClient (in production code, most of the test code has been ported to using RestAssured).

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
marcjohnson-kint: @bohdan-suprun Agreed, adding the UUID regular expression constraint to the interface will be a breaking change.

What does this change currently do, try to detect invalid UUIDs or not?

kaladay: The test in question is testing to confirm that an invalid uuid throws an invalid uuid (`LoanPoliciesApiTest.loanPolicyWithFixedDate`). This appears to be a test checking for a bad foreign key due to an invalid UUID being passed.

Its expecting this message: `Bad foreign key`, with a 422 response.

Instead, it gets a 500 response with `invalid input syntax for type uuid: "1234567890"` as the message.

To me like an invalid UUID is a bad foreign key. If the test is intended to not test that an invalid UUID could be a bad foreign key but instead to simply test for a non-existent UUID, then I believe the suggestion from @bohdan-suprun makes the most sense.

src/main/java/org/folio/rest/impl/RequestsAPI.java
kaladay: > Can the MyPgUtil use be replaced with PgUtil within this pull request?

It looks to me like doing that would incur some significant changes, possibly breaking changes. The MyPgUtil is being used only for the Upserts, which as far as I can tell were not ported over to PgUtil.

I can think of a few approaches here:
1. Implement upserts in PgUtil, but I imagine they didn't implement it for some reason.
1. Implement a select, check its response, and then conditionally call insert/update as appropriate.
1. Redesign something to avoid the need for an upsert, if this is even possible.
1. It is already marked as deprecated, some parts of these tests could be somehow marked as requiring changes relating to the deprecating functions (commenting out affected tests or implementing temporary solutions, for example).

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
kaladay: > What does this change currently do, try to detect invalid UUIDs or not?

The change currently tries to detect invalid UUIDs.

src/test/java/org/folio/rest/support/WebClient.java
kaladay: I feel like Okapi is completely designed around the HTTP protocol.
Would it be fine with shortening `OkapiHttpClient` to `OkapiClient`, or is that too generic?
Either way, I agree that using `OkapiHttpClient` or `OkapiClient` is an improvement over `HttpClient` or `WebClient`.

src/main/java/org/folio/rest/impl/LoansAPI.java
kaladay: > Can you expand upon what you mean here? Do you mean the error should be reported differently from PgUtil?

I was thinking more along the lines of changing the tests to look specifically for their table:value (aka: leave PgUtil alone).
Now that I am re-looking at the error message `value already exists in table loan: 335b7118-76e7-4d0a-9453-3dda29dfe529` is not that much more helpful. 

The error message `...value already exists in table loan:...` (HTTP 422) does not contain the unique index, and so there is no way to truly infer this.

The error message `...duplicate key value violates unique constraint "loan_itemid_idx_unique"...` (HTTP 400) does contain the index and can be inferred.

The other error message seems to be with `LoansPolicyApi.java` and not `LoansApi.java`.
For reference, that error message is `...invalid input syntax for type uuid: "null")...` (HTTP 500).

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
marcjohnson-kint: @kaladay What caused this behaviour to change during the upgrade?

src/main/java/org/folio/rest/impl/LoansAPI.java
marcjohnson-kint: > > Can you expand upon what you mean here? Do you mean the error should be reported differently from PgUtil?
> 
> I was thinking more along the lines of changing the tests to look specifically for their table:value (aka: leave PgUtil alone).

I'm ok with the tests changing to describe what the new error message is. We might not be able to retain the previous message.

> Now that I am re-looking at the error message `value already exists in table loan: 335b7118-76e7-4d0a-9453-3dda29dfe529` is not that much more helpful.

Is this a PostgreSQL error that is forwarded on by PgUtil or is PgUtil translating these error messages?

> The error message `...value already exists in table loan:...` (HTTP 422) does not contain the unique index, and so there is no way to truly infer this.
> 
> The error message `...duplicate key value violates unique constraint "loan_itemid_idx_unique"...` (HTTP 400) does contain the index and can be inferred.

Ok, so there are some cases where the error contains enough to figure out the specific cause?

src/main/java/org/folio/rest/impl/RequestsAPI.java
marcjohnson-kint: > > Can the MyPgUtil use be replaced with PgUtil within this pull request?
> 
> It looks to me like doing that would incur some significant changes, possibly breaking changes. The MyPgUtil is being used only for the Upserts, which as far as I can tell were not ported over to PgUtil.
> 
> I can think of a few approaches here:
> 
> 1. Implement upserts in PgUtil, but I imagine they didn't implement it for some reason.

Please can you ask the person who deprecated the upsert method in MyPgUtil what the recommended alternative approach is

> 2. Implement a select, check its response, and then conditionally call insert/update as appropriate.

That is what has been done in other places.

> 3. Redesign something to avoid the need for an upsert, if this is even possible.

I don't know enough about the context to answer that.

> 4. It is already marked as deprecated, some parts of these tests could be somehow marked as requiring changes relating to the deprecating functions (commenting out affected tests or implementing temporary solutions, for example).

Is MyPgUtil part of the production or test code?

If it is part of the production code, then are you proposing changing the tests to reflect changes to error message?

src/test/java/org/folio/rest/support/WebClient.java
marcjohnson-kint: > I feel like Okapi is completely designed around the HTTP protocol.

> Would it be fine with shortening `OkapiHttpClient` to `OkapiClient`, or is that too generic?
> Either way, I agree that using `OkapiHttpClient` or `OkapiClient` is an improvement over `HttpClient` or `WebClient`.

I'm ok with either `OkapiClient` or `OkapiHttpClient`.

I tend to prefer more specific names and it sets the context for the method names.

src/main/java/org/folio/rest/impl/RequestsAPI.java
kaladay: > Is MyPgUtil part of the production or test code?

MyPgUtil is part of production (`org.folio.rest.persist.MyPgUtil`).

In case 4, I am proposing **temporarily** changing the test code to reflect changes.
This would, in theory, allow for a new ticket to be opened up so that the resolution of the deprecated Upsert (and related necessary changes discovered in this PR) can be performed in completely separate ticket, branch, and PR.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
kaladay: @marcjohnson-kint  I didn't investigate into RMB. I only isolated the issue specifically to the upgrade from `26.2.3` into `27.1.0`, with a minimal set of required changes (removing `fullText` column from `schema.json` and replacing `HttpStatus.HTTP_VALIDATION_ERROR.toInt())` with `422`).

Glancing at the diff in RMB between tags `v26.2.3` and `v27.1.0`, specifically for `domain-models-runtime`, the following seem likely places to start investigating:
- There is a brand new class `PgExceptionFacade.java`.
- The PgUtil also adds `422` responses as well as new validation pattern matchers.

Investigating the cause for that behavioral change in RMB seems to me to be out of the scope of this issue.

src/main/java/org/folio/rest/impl/LoansAPI.java
kaladay: > Is this a PostgreSQL error that is forwarded on by PgUtil or is PgUtil translating these error messages?

This is a message _created by PgUtil_, see the link in my earlier comments (`PgUtil.java#L332`). The `responseUniqueViolation` method. Now that I am glancing at that method, it makes me wonder if that code is somehow failing. There is a regex match being used and if that regex fails to match, then it returns the custom generated response with either a 422 or a non-422 status.

> Ok, so there are some cases where the error contains enough to figure out the specific cause?

Yes and considered my above comment in regards to `responseUniqueViolation`, I am wondering if this happens in cases where the regex matches as intended.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
marcjohnson-kint: > @marcjohnson-kint I didn't investigate into RMB. I only isolated the issue specifically to the upgrade from `26.2.3` into `27.1.0`, with a minimal set of required changes (removing `fullText` column from `schema.json` and replacing `HttpStatus.HTTP_VALIDATION_ERROR.toInt())` with `422`).

Hmm, are they not the same?

> Glancing at the diff in RMB between tags `v26.2.3` and `v27.1.0`, specifically for `domain-models-runtime`, the following seem likely places to start investigating:
> 
> * There is a brand new class `PgExceptionFacade.java`.
> * The PgUtil also adds `422` responses as well as new validation pattern matchers.
> 
> Investigating the cause for that behavioral change in RMB seems to me to be out of the scope of this issue.

Agreed, I think the minimum amount of changes should be made to perform the upgrade and preserve as much of the existing behaviour as possible.

I think we should consider changes to messages, not to be part of the preservation, and adapt the tests to reflect the messages produced (as long as they can be done consistently and we think they are triggered for the right reasons).

And then raise separate technical debt issues for improving some of these areas in future. Which could include investigating RAML Module Builder more.

src/main/java/org/folio/rest/impl/LoansAPI.java
marcjohnson-kint: > > Is this a PostgreSQL error that is forwarded on by PgUtil or is PgUtil translating these error messages?
> 
> This is a message _created by PgUtil_, see the link in my earlier comments (`PgUtil.java#L332`). The `responseUniqueViolation` method. Now that I am glancing at that method, it makes me wonder if that code is somehow failing. There is a regex match being used and if that regex fails to match, then it returns the custom generated response with either a 422 or a non-422 status.

If you believe something is failing for the wrong reasons that should be investigated.

If it is likely to be only a change in message, then I agree, we should just change the expected message in the tests so they accurately describe what is happening.

src/main/java/org/folio/rest/impl/RequestsAPI.java
marcjohnson-kint: > > Is MyPgUtil part of the production or test code?
> 
> MyPgUtil is part of production (`org.folio.rest.persist.MyPgUtil`).
> 
> In case 4, I am proposing **temporarily** changing the test code to reflect changes.
> This would, in theory, allow for a new ticket to be opened up so that the resolution of the deprecated Upsert (and related necessary changes discovered in this PR) can be performed in completely separate ticket, branch, and PR.

Agreed, let's create new JIRA issue for the technical debt and change the test code where the behaviour remains intact yet the message has changed.

src/test/java/org/folio/rest/support/WebClient.java
```diff
@@ -53,34 +55,34 @@ public void post(URL url,
...
-      String encodedBody = Json.encodePrettily(body);
-      log.info(String.format("POST %s, Request: %s",
-        url.toString(), encodedBody));
-      request.end(encodedBody);
+
+    if (body == null) {
+      request.send(responseHandler);
     } else {
-      request.end();
+      Buffer encodedBody = Buffer.buffer(Json.encodePrettily(body));
```
julianladisch: Using the guard clause pattern you can avoid the else clause:

    if (body == null) {
      request.send(responseHandler);
      return;
    }
    Buffer encodedBody = Buffer.buffer(Json.encodePrettily(body));
    ...

src/main/java/org/folio/rest/impl/RequestsAPI.java
julianladisch: The collection rtype assumes that a PUT is an update only, not an upsert, and should return 404 when the record doesn't exists: https://github.com/folio-org/raml/blob/raml1.0/rtypes/item-collection.raml#L65
https://s3.amazonaws.com/foliodocs/api/mod-circulation-storage/request-storage.html#request_storage_requests__requestid__put
I propose to change the upsert to an update to match the API. This is a bug fix and not a breaking API change. Clients can (and usually do) use the POST API for inserts.
This is the reason why there is no need to have an upsert method in PgUtil.

In addition the current upsert implementation returns 204 on insert, it should return 201. This wrong status code is the reason why I marked MyPgUtil as deprecated.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -97,12 +74,6 @@
...
-          "fieldName": "itemId",
-          "tOps": "ADD",
```
julianladisch: Do not remove the lines, change `"tOps": "ADD"` to `"tOps": "DELETE"` so that the index gets removed.
This schema.json doesn't contain "itemId" as "fieldName" in any "foreignKey" section. Therefore an index is needed for searching. Please create an "index" section and add this.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -31,33 +28,13 @@
...
           "fieldName": "status.name",
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
         }
       ],
-      "index": [
-        {
-          "fieldName": "id",
-          "tOps": "ADD",
```
julianladisch: Do not remove the lines, change `"tOps": "ADD"` to `"tOps": "DELETE"` so that the index gets removed.

src/main/resources/templates/db_scripts/schema.json
julianladisch: We need an index for "userId" and "itemId", but not as an "ginIndex" but as a simple btree index ("index" section).
Do not remove the lines, for "ginIndex" change "tOps": "ADD" to "tOps": "DELETE" so that the indexes get removed.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private OkapiResponseUtil() {
+    throw new UnsupportedOperationException("Do not instantiate");
+  }
+
+  public static String getErrorMessage(AsyncResult<Response> reply) {
+    String message = null;
+
+    if (reply.succeeded() && reply.result().getStatus() >= 400
+      && reply.result().getStatus() < 500 && reply.result().hasEntity()) {
```
bohdan-suprun: Should 500 also be considered, i.e. `<=500`?

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
```
bohdan-suprun:
```
if (errors.getErrors().size() > 0) {
  Error error = errors.getErrors().get(0);
  message = error.getMessage().toLowerCase();
}
```

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
kaladay: I changed it from < 600 because I thought we weren't handling 5xx issues.

The reason why it is no longer handling the 500 issue is due to the change in how it is handling the UUID test. The 500 appears to have only happened on invalid UUID, which was not what the test was for. When we write a new test, then the 500 will need to be handled.

Should this utility handle all of the 5xx codes, the 4xx an 500, or just the 4xx codes?

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
kaladay: We only ever expect one Error in the Errors list?

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
bohdan-suprun: Hmm... that's a good question. As far as I understand the method is mostly used to check the 'edge' cases, like same position in queue, constraint violation, etc., right? 

If so, then I think it is fine to use only 4xx, since you're expecting that constraint violation is a 4xx status code, and a 5xx code is most likely is not the case we're trying to handle.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
bohdan-suprun: I think yes. Alternatively we can concatenate all of them, or return an array. 

I just spotted that you're rewriting the `message` on each iteration, we can simplify it somehow.

kaladay: > I just spotted that you're rewriting the message on each iteration, we can simplify it somehow.

Oops, I must have forgotten that logic when I moved the `contains()` check outside of the loop.
As it stands, my logic there is unsound and needs to be fixed.
The concatination may be a good way to catch all of the possibilities.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: @julianladisch I thought there was a change near the end of 2019 Q4 where RAML Module Builder dropped all indexes during an upgrade and recreated only the remaining ones?

Or have I mis-recalled that situation, and a different path was chosen?

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
marcjohnson-kint: There is no guarantee that there will only be one error. We should likely forward on any errors we receive from PgUtil

src/test/java/org/folio/rest/support/WebClient.java
marcjohnson-kint: Thanks for renaming this.

src/test/java/org/folio/rest/api/CancellationReasonsApiTest.java
marcjohnson-kint: Thanks for introducing some custom matchers to make this more readable

src/main/java/org/folio/rest/impl/RequestsAPI.java
marcjohnson-kint: I think, where possible, we should preserve the current behaviour of mod-circulation-storage prior to this upgrade (excluding specific error messages)

src/main/resources/templates/db_scripts/schema.json
julianladisch: RMB drops all indexes listed in schema.json and then recreates them because most of them were changed for the 600 char length limit.
RMB does not touch indexes not listed in the indexes sections of schema.json. Module authors may use customSnippetPath to create special indexes.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,76 @@
...
+   *   The Okapi Response to parse for error messages.
+   * @param subString
+   *   The substring to find within the error messages.
+   *
+   * @return
+   *   TRUE is returned on match, FALSE otherwise.
+   */
+  public static boolean containsErrorMessage(AsyncResult<Response> reply, String subString) {
+    String message = getErrorMessage(reply);
+    return message != null && message.contains(subString);
```
julianladisch: The comparison was case sensitive before.
Is the change to have a case insensitive comparison intended?
If yes you should also lowercase subString: `message.contains(subString.toLowerCase())` and extend the javadoc of containsErrorMessage and getErrorMessage accodingly.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: @julianladisch Thank you for clarifying that distinction, it helps my understanding of the behaviour

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
kaladay: With postgresql being case-sensitive by design, I am concerned that the error messages could have different cases. I feel like we shouldn't care about the case. Lowercasing it was an attempt to avoid potential fragility in the messages. This may, however, be unecessary caution (and reflexive habit) on my part. I will remove the toLowerCase() and just make this project consistent. Changing the case sensitivity could be in a different issue/ticket.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    if (reply.succeeded() && reply.result().getStatus() >= 400
+      && reply.result().getStatus() < 500 && reply.result().hasEntity()) {
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        if (errors.getErrors().size() > 0)
```
bohdan-suprun:
```suggestion
        if (errors.getErrors().size() > 0) {
```

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,76 @@
...
+          message = "";
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
+          message += error.getMessage().toLowerCase() + "\n";
```
bohdan-suprun: I'd use StringBuilder here.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @param reply
+   *   The Okapi Response to parse for error messages.
+   *
+   * @return
+   *   A string of concatenated error messages associated with the reply.
+   *   A newline is appended at the end of each distinct error message.
+   */
+  public static String getErrorMessage(AsyncResult<Response> reply) {
+    String message = null;
```
julianladisch: Why do you use null? You can simply return an empty string if no error message is found in Response.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,75 @@
...
+   *   The Okapi Response to parse for error messages.
+   * @param subString
+   *   The substring to find within the error messages.
+   *
+   * @return
+   *   TRUE is returned on match, FALSE otherwise.
+   */
+  public static boolean containsErrorMessage(AsyncResult<Response> reply, String subString) {
+    String message = getErrorMessage(reply);
+    return message != null && message.contains(subString);
```
julianladisch: Can getErrorMessage been changed to never return null, to return an empty string ""?

src/test/java/org/folio/rest/api/CirculationRulesApiTest.java
```diff
@@ -57,7 +57,7 @@ private JsonResponse getResponse() throws Exception {
   /** @return the JSON of the get response, asserts an HTTP_OK=200 response */
   private JsonObject get() throws Exception {
     JsonResponse response = getResponse();
-    assertThat(response.getBody(), response.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(response.getBody(), response, matchesOk());
```
julianladisch:
```suggestion
    assertThat(response, matchesOk());
```
matchesOk = matchesSafely automatically appends response.getBody() to the failure message.

src/test/java/org/folio/rest/api/CirculationRulesApiTest.java
```diff
@@ -72,7 +72,7 @@ private JsonResponse putResponse(CirculationRules circulationRules) throws Excep
   /** @return the message of the put response, asserts an HTTP_OK=200 response */
   private String put204(CirculationRules circulationRules) throws Exception {
     JsonResponse response = putResponse(circulationRules);
-    assertThat(response.getBody(), response.getStatusCode(), is(HttpURLConnection.HTTP_NO_CONTENT));
+    assertThat(response.getBody(), response, matchesNoContent());
```
julianladisch:
```suggestion
    assertThat(response, matchesNoContent());
```
matchesSafely automatically appends response.getBody() to the failure message.

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
kaladay: This is just me being cautious.

https://github.com/folio-org/mod-circulation-storage/blob/ac1e03e575b2407e167654ddd0ee69350da5db26/src/main/java/org/folio/rest/impl/util/RequestsApiUtil.java#L23

This code is considering a NULL instead of an empty string, and so I was worried that there may be places I haven't noticed that might be considering, expecting, or using NULL.

I would prefer to use an empty string, then I could just remove the assignment condition (`if (errors.getErrors().size() > 0)`).

src/main/java/org/folio/rest/impl/util/OkapiResponseUtil.java
```diff
@@ -0,0 +1,74 @@
...
+      && reply.result().getStatus() < 500 && reply.result().hasEntity()) {
+
+      // When entity is an instance of Errors, the getEntity().toString()
+      // will return an object identifier. Process the object to correctly
+      // parse the message.
+      if (reply.result().getEntity() instanceof Errors) {
+        Errors errors = (Errors) reply.result().getEntity();
+
+        for (int i = 0; i < errors.getErrors().size(); i++) {
+          Error error = errors.getErrors().get(i);
```
julianladisch: You may reduce this to

    for (Error error : errors.getErrors()) {

marcjohnson-kint:
@julianladisch @bohdan-suprun What still needs to be done for this work to be approved?

bohdan-suprun:
> @julianladisch @bohdan-suprun What still needs to be done for this work to be approved?

I'm fine with the changes. There is a conflict in PR but I don't think it require additional code changes.

PR #233, opened 2020-02-03
CIRCSTORE-187: Add claimedReturnedDate property for loan.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-187: Backend: Claim returned: mark an item claim returned

Add claimed returned date for loan.

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -98,6 +98,11 @@
       "type" : "string",
       "format": "date-time"
     },
+    "claimedReturnedDate": {
+      "description": "Date and time the item was claimed returned for this loan",
+      "type": "string",
+      "format": "date-time"
+    },
     "overdueFinePolicyId": {
       "description": "ID of overdue fines policy at the time the item is check-in or renewed",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -15,7 +15,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "6.5",
+      "version": "6.6",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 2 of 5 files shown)


ramls/loan.json
```diff
@@ -112,6 +112,27 @@
...
+        "dateTime": {
+          "description": "Date and time the item was claimed returned for this loan",
+          "type": "string",
+          "format": "date-time"
+        },
+        "staffMemberId": {
+          "type": "string",
+          "description": "Staff member id who marked the loan as claimed returned",
+          "$ref": "raml-util/schemas/uuid.schema"
+        }
```
marcjohnson-kint: Why does the staff member who marked an item as claimed returned during this loan need to be recorded? Is there a scenario in the dependant stories that requires this?

bohdan-suprun: Yes, check scenario # 6:
> Given a loan for an item that is Claim returned
> * When viewing loan details
> * Then populate as follows: 
>    * Claimed returned: show date and time item was claimed returned
>    * Action history: show row for claimed returned
>      * ...
>      * Source: Lastname, firstname (and link to user profile) of user who marked item claim returned

marcjohnson-kint: @bohdan-suprun Thanks

As I understand it, the action history is based upon the audit tracking of the loans table in mod-circulation-storage. 

I think `Source` typically refers to the change metadata included in each record. Is that sufficient for this change, or are we expecting the UI to include special logic for the `claimed returned` action to use different information from the record?

Does that fit with your understanding?

bohdan-suprun: Thanks, @marcjohnson-kint. That's make sense. I think we have to check with Emma, how important the staff member id for this feature. Because it maybe a case when:
* User `A` marks the item as claimed returned: `updatedByUserId` is `A`
* Then user `B` updates/closes the loan, so the `updatedByUserId` is `B` now.

What do you think?

marcjohnson-kint: @bohdan-suprun I think checking with Emma is ok.

If I'm understanding scenario 6 correctly, this information is only relevant for action history, which will not be effected by subsequent changes to the loan because it is based on copied snapshots of the loan on every write.

Does that make sense?

bohdan-suprun: Aha, right. Will remove the `staffMemberId` then.

PR #235, opened 2020-02-13
CIRCSTORE-193: Implement check-in storage API

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-193: Store a check in.

New `/check-in-storage/check-ins` API has been implemented. There are following endpoints:
* `POST` `/check-in-storage/check-ins`;
* `GET` `/check-in-storage/check-ins` - get by query;
* `GET` `/check-in-storage/check-ins/{operationId}` - get record by id.

New DB table has been added `check_in`.


descriptors/ModuleDescriptor-template.json
```diff
@@ -416,9 +416,71 @@
...
+    },
+    {
+      "id": "check-in-operation-storage",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "GET"
+          ],
+          "pathPattern": "/check-in-operation-storage",
```
marcjohnson-kint: Typically, with the storage APIs, they use a structure like `/item-storage/items` could these be adapted to fit that kind of structure, e.g. `/check-in-operation-storage/check-ins`?

descriptors/ModuleDescriptor-template.json
```diff
@@ -416,9 +416,71 @@
           "permissionsRequired": ["patron-action-session-storage.expired-session-patron-ids.collection.get"]
         }
       ]
+    },
+    {
+      "id": "check-in-operation-storage",
```
marcjohnson-kint: I'm not keen on `operation`, yet I've been trying to to think of an alternative, without much success. What do you think of `check-in-storage`?

```suggestion
      "id": "check-in-operation-storage",
```

bohdan-suprun: Fine for me, do you mean to rename URL as well?

marcjohnson-kint: What I'm thinking is that the interface ID is `check-in-storage`, and the paths are something like `check-in-storage/check-ins`

marcjohnson-kint: This is assuming that FOLIO is still using plural path names for its endpoints

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+      "type": "string",
+      "description": "Date and time when the check-in has been occurred",
+      "format": "date-time"
+    },
+    "itemId": {
+      "type": "string",
+      "description": "Id of item that has checked in",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "checkInServicePointId": {
```
marcjohnson-kint: I think this can be `servicePointId` as it is in the context of a check in record. What do you think?

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique ID (generated UUID) of the check-in operation",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "occurredDateTime": {
+      "type": "string",
+      "description": "Date and time when the check-in has been occurred",
```
marcjohnson-kint:
```suggestion
      "description": "Date and time when the check-in occurred",
```

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Check-in operation",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique ID (generated UUID) of the check-in operation",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "occurredDateTime": {
```
marcjohnson-kint: I've been wondering about this, and whether we want a separate property for when the event was recorded.

What do you think?

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+      "format": "date-time"
+    },
+    "itemId": {
+      "type": "string",
+      "description": "Id of item that has checked in",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "checkInServicePointId": {
+      "type": "string",
+      "description": "Check-in service point id",
```
marcjohnson-kint:
```suggestion
      "description": "Service point where the item was checked in",
```

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "performedByUserId": {
+      "type": "string",
+      "description": "Id of user that has performed the check-in operation",
```
marcjohnson-kint:
```suggestion
      "description": "Id of user that has performed check-in operation",
```

ramls/check-in.json
```diff
@@ -0,0 +1,40 @@
...
+      "$ref": "raml-util/schemas/uuid.schema"
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "id",
```
marcjohnson-kint: Can `id` be required, if this JSON schema is used for the request body of the POST to create the record?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -258,6 +258,11 @@
       "tableName": "patron_action_session",
       "withMetadata": true,
       "withAuditing": false
+    },
+    {
+      "tableName": "check_in",
+      "withMetadata": true,
+      "withAuditing": false
```
marcjohnson-kint: Do tables need a `fromVersion`?

descriptors/ModuleDescriptor-template.json
```diff
@@ -416,9 +416,71 @@
...
+        },
+        {
+          "methods": [
+            "DELETE"
+          ],
+          "pathPattern": "/check-in-storage/check-ins/{id}",
+          "permissionsRequired": [
+            "check-in-storage.check-ins.item.delete"
+          ]
+        },
```
marcjohnson-kint: Why do we allow a check in event to be deleted?

ramls/check-in.json
bohdan-suprun: I think it should be important for reports, as Emma mentioned they wanted to be able to query logs for some time period, e.g. last month, year, etc.

Or do you mean something else?

marcjohnson-kint: Apologies, I suggesting the idea that the `check in` event has both `occuredAt` for when the check in occurred and `recordedAt` for when that was recorded by the system.

Now that I think about it, I think the `metadata.createdDate` might be sufficient for the latter.

ramls/check-in.json
bohdan-suprun: I decided to mark it is as required (actually PUT is not implemented now). 

There was discussion on an inventory PR that an endpoint has optional ID field, but at the same time it shares schema for POST and PUT and for PUT the `id` is required.

ramls/check-in.json
bohdan-suprun: Agree, the `metadata.createdDate` can be used as a `recordedAt` property.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: I was thinking that some time we will have a job that will remove "expired" records. But now I realized that a delete by a query will better suite here.

Should I remove this ability? What do you think?

ramls/check-in.json
bohdan-suprun: Renamed to `servicePointId`

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Will remove the endpoint, It should be easy enough to add once we need.

ramls/check-in.json
marcjohnson-kint: I don't think FOLIO has a clear standard for this, which makes it confusing.

Usually, the POST endpoint allows for the defaulting of the `id`, is that done for this API?

bohdan-suprun: No, the endpoint does not have defaulting for id.

marcjohnson-kint: I think we should be consistent and provide the ability to default the ID

(It really isn't needed for this API, though if once path were chosen, I'd pick always defaulting the id)

src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
```diff
@@ -0,0 +1,138 @@
...
+    checkInClient.create(secondCheckIn);
+    checkInClient.create(thirdCheckIn);
+
+    MultipleRecords<JsonObject> allCheckInOperations = checkInClient
+      .getMany(String.format("occurredDateTime >= %s", beforeCreateDateTime.getMillis()));
+
+    assertThat(allCheckInOperations.getTotalRecords(), is(3));
+    assertTrue(allCheckInOperations.getRecords().contains(firstCheckIn));
+    assertTrue(allCheckInOperations.getRecords().contains(secondCheckIn));
+    assertTrue(allCheckInOperations.getRecords().contains(thirdCheckIn));
```
marcjohnson-kint: Given this test is intended to demonstrate searching by when a check in occurred, could it demonstrate how they are filtered by only including some of the records in the expected response?

src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  @Test
+  public void cannotUseNegativeOffsetForSearch()
+    throws InterruptedException, ExecutionException, TimeoutException,
+    MalformedURLException {
+
+    JsonResponse jsonResponse = checkInClient
+      .attemptGetMany("occurredDateTime is not null", -10, 10);
```
marcjohnson-kint: Is `is not null` a valid CQL query?

bohdan-suprun: No, the test should fail before processing CQL by RMB.

marcjohnson-kint: I think it would better demonstrate the intent of this test if the CQL was valid.

That would also mean the test might not start failing if the RAML Module Builder changes the order in which it validates requests

src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    JsonObject checkInToCreate = createSampleCheckIn()
+      .withId(null)
+      .create();
+
+    IndividualResource createResult = checkInClient.create(checkInToCreate);
+
+    assertThat(createResult.getId(), notNullValue());
+
+    checkInToCreate.put("id", createResult.getId());
```
marcjohnson-kint: Why is an `id` added to the record after creation?

bohdan-suprun: In order to simplify comparison - compare json only instead of field by field.

marcjohnson-kint: I think that comes with a trade off between this reuse and the clarity of the tests. I think this obscures the intent that the `id` is server generated.

Maybe an alternative would be to to check the record (inc not null ID) except the ID value within the shared method and check the ID separately in the relevant tests.

bohdan-suprun: I've added matcher that will compare only those properties that present on both records.

src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
marcjohnson-kint: @bohdan-suprun Why was this comment resolved? The test appears to still expect all of the records be returned, which to me, does not demonstrate any filtering by the CQL query.

bohdan-suprun: I've added a separate test to verify this case, see `canFilterCheckInsByServicePointAndItem` below.

marcjohnson-kint: In which case, what is this test intended to demonstrate?

To me, a search that returns all of the records in a system does not demonstrate that the search did any filtering, what am I missing?

bohdan-suprun: I've merged those two tests.

PR #236, opened 2020-02-26
CIRCSTORE-196: Store additional check-in properties for in-house use.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-196: Store additional properties in check-in-storage.

Add properties needed for in-house use:
* itemStatus
* itemLocationId;
* requestQueueSize;

### DIFF
--- a/ramls/check-in.json
+++ b/ramls/check-in.json
@@ -18,9 +18,9 @@
       "description": "Id of item that has checked in",
       "$ref": "raml-util/schemas/uuid.schema"
     },
-    "itemStatus": {
+    "itemStatusPriorToCheckIn": {
       "type": "string",
-      "description": "Item status on check-in"
+      "description": "Item status prior to check-in"
     },
     "requestQueueSize": {
       "type": "integer",
@@ -29,7 +29,7 @@
     },
     "itemLocationId": {
       "type": "string",
-      "description": "Effective location of the item",
+      "description": "Location of the item in check-in time",
       "$ref": "raml-util/schemas/uuid.schema"
     },
     "servicePointId": {
--- a/ramls/check-in.json
+++ b/ramls/check-in.json
@@ -18,6 +18,20 @@
       "description": "Id of item that has checked in",
       "$ref": "raml-util/schemas/uuid.schema"
     },
+    "itemStatus": {
+      "type": "string",
+      "description": "Item status on check-in"
+    },
+    "requestQueueSize": {
+      "type": "integer",
+      "description": "Request queue size for the item",
+      "minimum": 0
+    },
+    "itemLocationId": {
+      "type": "string",
+      "description": "Effective location of the item",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
     "servicePointId": {
       "type": "string",
       "description": "Service point where the item was checked in",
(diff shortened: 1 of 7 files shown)


src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
```diff
@@ -171,12 +171,46 @@ public void cannotGetCheckInByIdIfDoesNotExists() throws InterruptedException,
...
+      .create();
+
+    JsonResponse createResult = checkInClient.attemptCreate(checkInToCreate);
+
+    assertThat(createResult,
+      isValidationResponseWhich(hasMessage("must be greater than or equal to 0")));
+    assertThat(createResult,
+      isValidationResponseWhich(hasParameter("requestQueueSize", "-1")));
+  }
+
```
Svitlana-Zmiivska: Does it make sense to create a test with non-existent location id?

bohdan-suprun: Sure, added.

bohdan-suprun:
@marc-johnson,

They all are optional, and it is allowed to create a record without them.

The existing positive test was updated and verifies them also.

ramls/check-in.json
```diff
@@ -18,6 +18,20 @@
       "description": "Id of item that has checked in",
       "$ref": "raml-util/schemas/uuid.schema"
     },
+    "itemStatus": {
```
marcjohnson-kint: I think it might be worth making it explicit that this is the item status prior to check in.

```suggestion
    "itemStatusPriorToCheckIn": {
```

ramls/check-in.json
```diff
@@ -18,6 +18,20 @@
...
+      "description": "Item status on check-in"
+    },
+    "requestQueueSize": {
+      "type": "integer",
+      "description": "Request queue size for the item",
+      "minimum": 0
+    },
+    "itemLocationId": {
+      "type": "string",
+      "description": "Effective location of the item",
```
marcjohnson-kint: I don't think the circulation domain is concerned with how the location for the item is determined, and so the description does not need to include `effective` to refer to the derived nature of the property in inventory.

src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
```diff
@@ -76,12 +76,15 @@ public void canCreateRecordWithoutId() throws InterruptedException,
...
-    JsonObject checkInToCreate = new CheckInBuilder().create();
+    UUID recordId = UUID.randomUUID();
+    JsonObject checkInToCreate = new CheckInBuilder().withId(recordId).create();
 
     JsonResponse createResponse = checkInClient.attemptCreate(checkInToCreate);
 
     assertThat(createResponse, isValidationResponseWhich(hasMessage("may not be null")));
     assertThat(createResponse, isValidationResponseWhich(hasParameter("occurredDateTime", "null")));
+    assertThat(checkInClient.attemptGetById(recordId).getStatusCode(),
+      is(404));
```
marcjohnson-kint: Why is this line split, when the lines above it are not?

marcjohnson-kint:
> @marcjohnson-kint,
> 
> They all are optional, and it is allowed to create a record without them.

Does that mean that it is acceptable for a check in to occur when this information is not known?

> The existing positive test was updated and verifies them also.

Based upon this pull request, the only test that has changed is a negative scenario (`cannotCreateCheckInIfRequiredPropertyMissing`). 

What am I missing? Is it because the builder has changed, and the checks in the tests are JSON equivalence?

bohdan-suprun:
@marcjohnson-kint,
 
> Does that mean that it is acceptable for a check in to occur when this information is not known?

I just thought that itemStatus, itemLocationId;, requestQueueSize properties can be looked up using the item ID, if it is not mandatory for a logging case. Do you think they should be required?

> Is it because the builder has changed, and the checks in the tests are JSON equivalence?

Yes, the JSON built by builder is get compared with received JSON.

PR #239, opened 2020-03-12
CIRCSTORE-199 Add billedLostItemFlag and dateBilledLostItem to loan s…

iselagb:
https://issues.folio.org/browse/CIRCSTORE-199

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -113,21 +113,27 @@
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
       "type": "string"
     },
-    "billedLostItemFlag": {
-      "description": "Indicates if the aged to lost fee has been billed (for use where delayed billing is set up)",
-      "type": "boolean"
-    },
-    "dateBillLostItem": {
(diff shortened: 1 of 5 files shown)


ramls/loan.json
```diff
@@ -113,6 +113,15 @@
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
       "type": "string"
     },
+    "billedLostItemFlag": {
```
bohdan-suprun: Can we remove the flag and treat `dateBillLostItem` presence as `true`.

ramls/loan.json
```diff
@@ -113,6 +113,15 @@
...
+      "description": "Indicates if the aged to lost fee has been billed (for use where delayed billing is set up)",
+      "type": "boolean"
+    },
+    "dateBillLostItem": {
```
bohdan-suprun: Is it possible to rename it?
```suggestion
    "billLostItemDate": {
```

marcjohnson-kint:
I think I'm lacking a significant amount of context on this work.

Why is this information needed on the loan? Is the information about when a fee / fine was charged not kept on the fee / fine itself? And that has a relationship to the loan?

PR #240, opened 2020-03-25
CIRCSTORE-201 Add markedMissingDate property to a loan

Svitlana-Zmiivska:
https://issues.folio.org/browse/CIRCSTORE-201 Claimed returned: mark an item missing

**Approach:** 
Add marked missing date for loan.

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -103,6 +103,11 @@
       "type": "string",
       "format": "date-time"
     },
+    "markedMissingDate": {
+      "description": "Date and time the item was marked missing for this loan",
+      "type": "string",
+      "format": "date-time"
+    },
     "overdueFinePolicyId": {
       "description": "ID of overdue fines policy at the time the item is check-in or renewed",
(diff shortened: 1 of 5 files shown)


ramls/loan.json
```diff
@@ -103,6 +103,11 @@
       "type": "string",
       "format": "date-time"
     },
+    "markedMissingDate": {
```
bohdan-suprun: Hi @Svitlana-Zmiivska, from what I last heard from @SergiyVSergiyenko, UI does not use neither `declaredLostDate` nor `claimedReturnedDate`. They use metadata.updatedDate from the loan history record.

It seems that we can omit this property.

@SergiyVSergiyenko, @zburke, @mkuklis am I correct understand?

SergiyVSergiyenko: @bohdan-suprun, this is not entirely true. To display the 'Claimed Return Date' value, we use `loan.claimedReturnedDate` property. And for the 'Declared Lost Date' value - really metadata.updatedDate (probably this will need to be fixed in the future).

bohdan-suprun: Aha, ok, thanks for clarifying this. We need the property then.

PR #241, opened 2020-03-25
Revert "CIRCSTORE-201 Add markedMissingDate property to a loan"

Svitlana-Zmiivska:
Reverts folio-org/mod-circulation-storage#240

**Reason:**
According to requirements, we don't need markedMissingDate property for a loan

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -15,7 +15,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "6.7",
+      "version": "6.6",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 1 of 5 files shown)


bohdan-suprun:
Hi @Svitlana-Zmiivska, 

Have you got a clarification from Emma, that this property is not needed?

As far as I understand **Action date** here:

image

means the property to be used.

However, loanActionHistory.metadata.updatedDate can be used here instead.

marcjohnson-kint:
@bohdan-suprun  As I understand it, loan actions come from the `loan action history` API, which includes a date and time when that snapshot of the loan record was taken. I believe that is what the UI interprets as `action date`.

Otherwise, it would need to choose different properties from the loan record depending upon the action, which I don't think it does.

Does that make sense?

bohdan-suprun:
@marcjohnson-kint, yes, make sense to me. Thank you!

Svitlana-Zmiivska:
@bohdan-suprun @marc-johnson This property isn't used on loan actions.
@SergiyVSergiyenko thank you for your check

PR #242, opened 2020-04-10
CIRCSTORE-208: rename itemEffectiveLocationAtCheckOut

rjruizes:
https://issues.folio.org/browse/CIRCSTORE-208

Follow-up to [CIRCSTORE-157](https://github.com/folio-org/mod-circulation-storage/pull/194) which added a new field itemEffectiveLocationAtCheckOut. This field name was renamed in mod-circulation; this issue is to propagate that rename to mod-circulation-storage.

### DIFF
--- a/ramls/examples/loan.json
+++ b/ramls/examples/loan.json
@@ -5,7 +5,7 @@
   "itemId": "cb20f34f-b773-462f-a091-b233cc96b9e6",
   "loanDate": "2017-03-01T23:11:00-01:00",
   "dueDate": "2017-03-15T23:11:00-01:00",
-  "itemEffectiveLocationAtCheckOut": "07f83048-6737-4c6a-a3bc-c49cd9beaa38",
+  "itemEffectiveLocationIdAtCheckOut": "07f83048-6737-4c6a-a3bc-c49cd9beaa38",
   "checkoutServicePointId": "8bb53832-c9fb-47ef-986d-5b1a498a8b30",
   "checkinServicePointId": "8bb53832-c9fb-47ef-986d-5b1a498a8b30",
   "status": {
--- a/ramls/loan-storage.raml
+++ b/ramls/loan-storage.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Loan Storage
-version: v6.7
+version: v6.8
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost:9130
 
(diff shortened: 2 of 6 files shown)


ramls/loan.json
```diff
@@ -21,7 +21,7 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
-    "itemEffectiveLocationAtCheckOut": {
+    "itemEffectiveLocationIdAtCheckOut": {
       "description": "The effective location, at the time of checkout, of the item loaned to the patron.",
       "type": "string",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
```
bohdan-suprun, line 27:
```suggestion
      "$ref": "raml-util/schemas/uuid.schema"
```

ramls/loan-storage.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Loan Storage
-version: v6.7
+version: v6.8
```
bohdan-suprun, line 3: Please, update module descriptor API version as well.

ramls/examples/loan.json
```diff
@@ -5,7 +5,7 @@
   "itemId": "cb20f34f-b773-462f-a091-b233cc96b9e6",
   "loanDate": "2017-03-01T23:11:00-01:00",
   "dueDate": "2017-03-15T23:11:00-01:00",
-  "itemEffectiveLocationAtCheckOut": "07f83048-6737-4c6a-a3bc-c49cd9beaa38",
+  "itemEffectiveLocationIdAtCheckOut": "07f83048-6737-4c6a-a3bc-c49cd9beaa38",
```
marcjohnson-kint, line 8: This is a breaking compatibility change, the `loan-storage` interface will need changing to version `7.0` and all dependent modules will need updating for compatibility.

marcjohnson-kint:
Thank you for updating this pull request to reflect the breaking change.

This change needs to be coordinated with changes to (at least) the following modules:
* ui-users
* mod-circulation
* stripes-core ([it appears](https://github.com/folio-org/stripes-core/blob/10bfbcf90c49cdb1244735147599f0313a784162/src/components/MainNav/AppList/tests/selectedApp.json#L14-L25))

There might also be other dependents like the API, regression or performance tests and reporting (which I am assuming you are aware of)

PR #244, opened 2020-04-17
[CIRCSTORE-207] Update schemas to support fee/fine scheduled notices

OleksandrVidinieiev:
To support implementation of [CIRC-711](https://issues.folio.org/browse/CIRC-711).

Resolves: [CIRCSTORE-207](https://issues.folio.org/browse/CIRCSTORE-207)
Original UI story: [UICIRC-451](https://issues.folio.org/browse/UICIRC-451)

### DIFF
--- a/ramls/patron-notice-policy.json
+++ b/ramls/patron-notice-policy.json
@@ -92,61 +92,11 @@
               },
               "sendBy": {
                 "type": "object",
-                "additionalProperties": false,
-                "description": "Send by options",
-                "properties": {
-                  "duration": {
-                    "type": "integer",
-                    "description": "Duration amount",
-                    "minimum": 1,
-                    "maximum": 100
-                  },
-                  "intervalId": {
-                    "type": "string",
-                    "description": "Interval id",
-                    "enum": [
-                      "Minutes",
-                      "Hours",
-                      "Days",
-                      "Weeks",
-                      "Months",
-                      "Years"
-                    ]
-                  }
-                },
-                "required": [
-                  "duration",
-                  "intervalId"
-                ]
+                "$ref": "interval.json"
               },
               "sendEvery": {
                 "type": "object",
-                "additionalProperties": false,
-                "description": "Send every options",
-                "properties": {
-                  "duration": {
-                    "type": "integer",
-                    "description": "Duration amount",
-                    "minimum": 1,
-                    "maximum": 100
-                  },
-                  "intervalId": {
-                    "type": "string",
-                    "description": "Interval id",
-                    "enum": [
-                      "Minutes",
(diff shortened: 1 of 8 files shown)


ramls/patron-notice-policy.json
```diff
@@ -207,74 +157,31 @@
             "properties": {
               "sendHow": {
                 "type": "string",
-                "description": "Defines how notice should be sent, before, after or upon",
+                "description": "Defines how notice should be sent, after or upon",
```
alexanderkurash, line 160: There should probably be dash or colon after "sent", otherwise it's a bit confusing.

ramls/patron-notice-policy.json
```diff
@@ -207,74 +157,31 @@
...
                 "enum": [
                   "Upon At",
-                  "Before",
```
alexanderkurash, line 213: Why are we removing this option?

OleksandrVidinieiev: Because this option makes no sense in terms of automated fees/fines.

ramls/patron-notice-policy.json
```diff
@@ -207,74 +157,31 @@
...
-                "required": [
-                  "duration",
-                  "intervalId"
-                ]
+                "$ref": "interval.json"
               }
             },
             "required": [
-              "sendHow"
+              "sendWhen"
```
alexanderkurash, line 184: I want to understand this change better. "how" is now not required, but "when" is. At the same time, "when" only has options for "Overdue fine returned" and "Overdue fine renewed". What will happen to other fee/fine notices? Are there other fee/fine notices (f.e. for manual fines)?

ramls/patron-notice-policy.json
alexanderkurash: So this section is only for automated fees/fines?

ramls/patron-notice-policy.json
OleksandrVidinieiev: This is the way it is in other two notice categories in this schema, `loanNotices` and `requestNotices`. This change will not affect any existing notices, since there are none. Manual fees/fines are created in `mod-feesfines` regardless of patron notice policy.

alexanderkurash: Thanks

ramls/patron-notice-policy.json
OleksandrVidinieiev: Yes, at least at the moment.

ramls/patron-notice-policy.json
OleksandrVidinieiev: Fixed.

PR #248, opened 2020-06-17
CIRCSTORE-215: Add indexes for request and scheduled_notice tables

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-215 - Missing Indexes Warnings.

### Changes:
* add indexes needed for `request` table;
* Add indexes needed for scheduled notices timer API, which sends query `nextRunTime<"now()" and noticeConfig.sendInRealTime=="true" and triggeringEvent==("an-event") requestId=="a-request-id" sortBy nextRunTime/sort.ascending`;
* Add tests dependencies on: `lombok`, `spring-testing/context/beans`, `restassured`;
* Use Spring DI capabilities for tests: ShceduledNotices and RequestApi


src/test/java/org/folio/rest/support/clients/CqlQuery.java
```diff
@@ -0,0 +1,29 @@
...
+  public static CqlQuery startsWith(String propertyName, Object value) {
+    return fromTemplate("%s=\"%s\"", propertyName, value);
+  }
+
+  public static CqlQuery lessThen(String propertyName, Object bound) {
+    return fromTemplate("%s<\"%s\"", propertyName, bound);
+  }
+
+  public static CqlQuery fromTemplate(String template, Object... values) {
+    return new CqlQuery(String.format(template, values));
```
julianladisch, line 23: Backslash \ and double quotes " need to be masked by prepending a backslash.
CQL quoted string: https://www.loc.gov/standards/sru/cql/spec.html#bnf

bohdan-suprun: Thanks, done.

src/test/java/org/folio/rest/support/clients/CqlQuery.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  private static Object maskCqlValue(Object value) {
+    if (!(value instanceof String) || StringUtils.isBlank((String) value)) {
+      return value;
+    }
+
+    return value.toString().replace("\\", "\\\\")
+      .replace("\"", "\\\"");
+  }
```
julianladisch, line 44: Can you mask all cases? Look at these tests:

```
package org.folio.rest.support.clients;

import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.assertThat;

import org.junit.Test;

public final class CqlQueryTest {
  @Test
  public void backslashString() {
    assertThat(CqlQuery.exactMatch("foo", "\\").asString(), is("foo==\"\\\\\""));
  }

  @Test
  public void backslashChar() {
    assertThat(CqlQuery.exactMatch("foo", '\\').asString(), is("foo==\"\\\\\""));
  }

  @Test
  public void backslashClass() {
    class MyClass {
      public String toString() {
        return "\\";
      }
    }
    assertThat(CqlQuery.exactMatch("foo", new MyClass()).asString(), is("foo==\"\\\\\""));
  }

  // If %c is supported is must properly mask.
  // Removing support for %c is fine.
  @Test
  public void charFromInt() {
    int c = '\\';
    assertThat(CqlQuery.fromTemplate("foo==\"%c\"", c).asString(), is("foo==\"\\\\\""));
  }
}
```

My suggestion is to remove support for %c and do it this way:

```
  private static String maskCqlValue(Object value) {
    return value.toString()
        .replace("\\", "\\\\")
        .replace("\"", "\\\"")
        .replace("*", "\\*")
        .replace("?", "\\?")
        .replace("^", "\\^");
  }
```

This also masks * ? ^ that are special CQL characters: https://www.loc.gov/standards/sru/cql/contextSets/theCqlContextSet.html

bohdan-suprun: Sure, modified.

src/test/java/org/folio/rest/support/matchers/CollectionMatchers.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+import io.vertx.core.json.JsonObject;
+
+public final class CollectionMatchers {
```
marcjohnson-kint, line 16: Was this class sourced from somewhere else?

src/test/java/org/folio/rest/api/RequestsApiTest.java
```diff
@@ -36,6 +41,7 @@
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
+import org.folio.rest.jaxrs.model.Request;
```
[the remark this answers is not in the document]
julianladisch: RMB uses `ramls/request.json` to generate `target/generated-sources/raml-jaxrs/org/folio/rest/jaxrs/model/Request.java` in the maven compile phase.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -167,6 +167,24 @@
...
+        {
+          "fieldName": "requesterId",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        }
+      ],
+      "fullTextIndex": [
+        {
+          "fieldName": "status",
```
julianladisch, line 186: The "ginIndex" for status should be removed.
We probably need an b-tree "index", not a "fullTextIndex" for status, see https://issues.folio.org/browse/CIRCSTORE-215

src/test/java/org/folio/rest/support/clients/CqlQuery.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public final String asString() {
+    return query;
+  }
+
+  private static Object maskCqlValue(Object value) {
+    if (!(value instanceof String) || StringUtils.isBlank((String) value)) {
+      return value;
+    }
```
julianladisch, line 40: The 2nd, 3rd and 4th test posted in my previous comment still fail. You can fix it by removing this if clause.

bohdan-suprun: Hi @julianladisch, this class is used for tests only. I don't think it make sense to invest a lot of time into this.

bohdan-suprun:
Hi @marcjohnson-kint, I don't use generated classes anymore, instead I declared separate ones for tests.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: I'm happy to remove them, thanks! The GIN index is marked as `DELETE` and a b-tree index is added:
```
{
    "fieldName": "status",
    "tOps": "ADD",
    "caseSensitive": false,
    "removeAccents": false
}
```

src/test/java/org/folio/rest/support/matchers/CollectionMatchers.java
bohdan-suprun: No, it is removed now.

PR #249, opened 2020-07-09
CIRCSTORE-219: Create request.requestType b-tree index

julianladisch:
request.requestType is an enum with these values: ["Hold", "Recall", "Page"]
The correct match operator is the field matching `==`
The existing "ginIndex" should be removed and a b-tree "index" created.
It should be case insensitive (for usability, the same as other request
b-tree indexes) and accents sensitive (we don't expect "Höld" searches.
The new index will also provide better totalRecords estimations resolving
https://issues.folio.org/browse/CIRCSTORE-210 .

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -150,7 +150,7 @@
         },
         {
           "fieldName": "requestType",
-          "tOps": "ADD",
+          "tOps": "DELETE",
           "caseSensitive": false,
           "removeAccents": true
         },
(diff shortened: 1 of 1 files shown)


julianladisch:
Front-end has already switche to `==`: https://github.com/folio-org/ui-requests/pull/605/files

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -180,6 +180,12 @@
           "caseSensitive": false,
           "removeAccents": false
         },
+        {
+          "fieldName": "requestType",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
```
marcjohnson-kint, line 187: Are these the settings that all b-tree indexes used for `==` should use now?

julianladisch: RMB has been changed to always use the accent and case setting of the underlying index.
The module can use the setting matching the front-end user needs.

PR #250, opened 2020-07-13
CIRCSTORE-222: Replace LoanPolicy Drools .drl rules by plain Java code

julianladisch:
RMB v30.0.1 ( https://issues.folio.org/browse/RMB-525 ) removes
RestVerticle's automatic Drools rules validation because Drools
is a big library and mod-circulation-storage is the only module
that uses it.

It is possible to put the drools code that was deleted in RMB into
mod-circulation-storage: https://github.com/folio-org/raml-module-builder/pull/672/files
However, the rules are very simple:
https://github.com/folio-org/mod-circulation-storage/tree/v12.0.0/src/main/resources/rules
Using Drools is over-engineering because a few Java if-clauses
suffice and converting the Drools rules to Java is trivial.


src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
```diff
@@ -190,6 +191,19 @@ public void loanPolicyWithFixedDate() throws
...
+        .put("name", "9").put("loanable", false).put("renewable", false)
+        .put("loansPolicy", new JsonObject());
+    CompletableFuture<TextResponse> completed9 = new CompletableFuture<>();
+    client.post(loanPolicyStorageUrl(),
+      loanPolicyRequest9, StorageTestSuite.TENANT_ID,
+      ResponseHandler.text(completed9));
+    TextResponse response9 = completed9.get(5, TimeUnit.SECONDS);
+    assertThat("loanable==false but loansPolicy exist", response9, matchesUnprocessableEntity());
+    // TODO: consider returning a better message
+    assertThat(response9.getBody(), containsString("Fixed due date"));
```
marcjohnson-kint, line 206: Please could this be defined as a separate test method rather than hidden amongst these other tests in the same method.

I think it would be useful, given that this code is changing to have a test demonstrating each of the four validations.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -110,7 +113,103 @@ public void putLoanPolicyStorageLoanPoliciesByLoanPolicyId(
...
+   * @return true if bool is TRUE, false if bool is FALSE or null
+   */
+  private boolean is(Boolean bool) {
+    return Boolean.TRUE.equals(bool);
+  }
+
+  void validate(LoanPolicy loanPolicy, Handler<AsyncResult<Response>> asyncResultHandler, Runnable runIfValid) {
+    final boolean isFixed = loanPolicy.getLoansPolicy() != null &&
+                            loanPolicy.getLoansPolicy().getProfileId() != null &&
+                            loanPolicy.getLoansPolicy().getProfileId().equalsIgnoreCase("Fixed");
```
marcjohnson-kint, line 137: Please can this use the standard two space line continuation indentation rather than horizontal alignment (which is harder to maintain and uses valuable horizontal space)

If there are concerns that this could make this less readable, maybe it could be extracted to a method which would reduce the indentation disparity and raise the abstraction of the current method.

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -110,7 +113,103 @@ public void putLoanPolicyStorageLoanPoliciesByLoanPolicyId(
     Handler<AsyncResult<Response>> asyncResultHandler,
     Context vertxContext) {
 
-    MyPgUtil.putUpsert204(LOAN_POLICY_TABLE, entity, loanPolicyId, okapiHeaders, vertxContext,
-        PutLoanPolicyStorageLoanPoliciesByLoanPolicyIdResponse.class, asyncResultHandler);
+    validate(entity, asyncResultHandler, () ->
+      MyPgUtil.putUpsert204(LOAN_POLICY_TABLE, entity, loanPolicyId, okapiHeaders, vertxContext,
+          PutLoanPolicyStorageLoanPoliciesByLoanPolicyIdResponse.class, asyncResultHandler));
```
marcjohnson-kint, line 118: Personally, I would prefer that the validate method returned something that allowed the process to continue or report an error, as that would allow the validate method to be organised differently and make the steps clearer. I understand that is challenging to do in this code at the moment.

Alternatively, there could be two parameters, one for what to do when valid and the other for what to do when invalid, which might be a consumer of an error response.

src/test/java/org/folio/rest/api/LoanPoliciesApiTest.java
julianladisch: cannotUpdateWhenFixedIsRenewable() and loanPolicyWithFixedDate() check the 1st validation.
cannotCreateLoanPolicyWithLoanableFalseAndLoansPolicy() and loanPolicyWithFixedDate() check the 2nd validation.
cannotUseHoldAlternateRenewalLoanPeriodForFixedProfile() checks the 3rd validation.
cannotUseRenewalsPeriodForFixedProfile() checks the 4th validation.

Sonar reports that all four rules are covered by tests:
https://sonarcloud.io/component_measures?id=org.folio%3Amod-circulation-storage&metric=new_coverage&pullRequest=250&selected=org.folio%3Amod-circulation-storage%3Asrc%2Fmain%2Fjava%2Forg%2Ffolio%2Frest%2Fimpl%2FLoanPoliciesAPI.java&view=list

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
julianladisch: Done

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -110,7 +115,107 @@ public void putLoanPolicyStorageLoanPoliciesByLoanPolicyId(
...
+        ||
+        ( loanPolicy.getRenewalsPolicy() != null
+          &&
+          ( ! isTrue(loanPolicy.getRenewable()) ||
+            ! isTrue(loanPolicy.getRenewalsPolicy().getDifferentPeriod())
+          )
+          && loanPolicy.getRenewalsPolicy().getAlternateFixedDueDateScheduleId() != null
+       )) {
+      String message =
+          "Alternate fixed due date cannot be " + loanPolicy.getRenewalsPolicy().getAlternateFixedDueDateScheduleId() +
```
craigmcnally, line 157: StringBuilder?

julianladisch: `javap -c LoanPoliciesAPI.class`
```
     128: new           #121                // class java/lang/StringBuilder
     131: dup
     132: ldc           #226                // String Alternate fixed due date cannot be
     134: invokespecial #228                // Method java/lang/StringBuilder."<init>":(Ljava/lang/String;)V
     137: aload_1
     138: invokevirtual #214                // Method org/folio/rest/jaxrs/model/LoanPolicy.getRenewalsPolicy:()Lorg/folio/rest/jaxrs/model/RenewalsPolicy;
     141: invokevirtual #223                // Method org/folio/rest/jaxrs/model/RenewalsPolicy.getAlternateFixedDueDateScheduleId:()Ljava/lang/String;
     144: invokevirtual #124                // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
     147: ldc           #231                // String  if renewable is
     149: invokevirtual #124                // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
     152: aload_1
     153: invokevirtual #208                // Method org/folio/rest/jaxrs/model/LoanPolicy.getRenewable:()Ljava/lang/Boolean;
     156: invokevirtual #233                // Method java/lang/StringBuilder.append:(Ljava/lang/Object;)Ljava/lang/StringBuilder;
     159: ldc           #236                // String , different period is
     161: invokevirtual #124                // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
     164: aload_1
     165: invokevirtual #214                // Method org/folio/rest/jaxrs/model/LoanPolicy.getRenewalsPolicy:()Lorg/folio/rest/jaxrs/model/RenewalsPolicy;
     168: invokevirtual #218                // Method org/folio/rest/jaxrs/model/RenewalsPolicy.getDifferentPeriod:()Ljava/lang/Boolean;
     171: invokevirtual #233                // Method java/lang/StringBuilder.append:(Ljava/lang/Object;)Ljava/lang/StringBuilder;
     174: ldc           #238                // String  and profile is
     176: invokevirtual #124                // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
     179: aload_1
     180: invokevirtual #193                // Method org/folio/rest/jaxrs/model/LoanPolicy.getLoansPolicy:()Lorg/folio/rest/jaxrs/model/LoansPolicy;
     183: invokevirtual #197                // Method org/folio/rest/jaxrs/model/LoansPolicy.getProfileId:()Ljava/lang/String;
     186: invokevirtual #124                // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
     189: invokevirtual #128                // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
     192: astore        5
     194: aload_0
     195: ldc           #240                // String alternateFixedDueDateScheduleId
     197: aload_1
     198: invokevirtual #214                // Method org/folio/rest/jaxrs/model/LoanPolicy.getRenewalsPolicy:()Lorg/folio/rest/jaxrs/model/RenewalsPolicy;
     201: invokevirtual #223                // Method org/folio/rest/jaxrs/model/RenewalsPolicy.getAlternateFixedDueDateScheduleId:()Ljava/lang/String;
```
https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

marcjohnson-kint: @julianladisch Please could you provide some context to that output in your response, so folks can follow the reasoning.

julianladisch: The Java compiler automatically converts the `+` string concatenation into StringBuilder.append invocations in cases like this.
As `+` is more readable than StringBuilder.append the `+` string concatenation should stay in the code.

marcjohnson-kint: Thanks @julianladisch I did not know that, and the explanation helps me understand better than the output and the link

craigmcnally: I think he's trying to say that the compiler will use the StringBuilder, so no need to make the code less readable

PR #251, opened 2020-07-13
 CIRCSTORE-220: Upgrade to raml-module-builder (RMB) 30.2.4 

julianladisch:
Updating RMB from 30.0.1 to RMB 30.2.4 fixes
https://issues.folio.org/browse/CIRCSTORE-210 "Requests -Filtering
by Request Type returns (Total Records = 1000) when there are more"
because it includes these fixes:

* [RMB-637](https://issues.folio.org/browse/RMB-637) Run ANALYZE after index (re-)creation
* [RMB-645](https://issues.folio.org/browse/RMB-645) Use where-only clause for the "count query"
* [RMB-639](https://issues.folio.org/browse/RMB-639) Unexpected breaking changes streamGet and getWithOptimizedSql

https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-302 advises:
> https://issues.folio.org/browse/RMB-652 error message "may not be null" changed to "must not be null" (hibernate-validator)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,8 +99,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <ramlfiles_path>${basedir}/ramls</ramlfiles_path>
-    <raml-module-builder-version>30.0.1</raml-module-builder-version>
-    <vertx-version>3.9.0</vertx-version>
+    <raml-module-builder-version>30.2.4</raml-module-builder-version>
+    <vertx-version>3.9.1</vertx-version>
     <lombok.version>1.18.12</lombok.version>
     <spring.version>5.2.7.RELEASE</spring.version>
     <rest-assured.version>3.3.0</rest-assured.version>
--- a/src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
+++ b/src/test/java/org/folio/rest/api/CheckInStorageApiTest.java
@@ -81,7 +81,7 @@ public class CheckInStorageApiTest extends ApiTests {
 
     JsonResponse createResponse = checkInClient.attemptCreate(checkInToCreate);
 
-    assertThat(createResponse, isValidationResponseWhich(hasMessage("may not be null")));
+    assertThat(createResponse, isValidationResponseWhich(hasMessage("must not be null")));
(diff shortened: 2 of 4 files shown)


julianladisch:
This pull request is blocked by merge of #250

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -110,7 +113,103 @@ public void putLoanPolicyStorageLoanPoliciesByLoanPolicyId(
...
+        )
+        ||
+        ( loanPolicy.getRenewalsPolicy() != null
+          &&
+          ( ! is(loanPolicy.getRenewable()) ||
+            ! is(loanPolicy.getRenewalsPolicy().getDifferentPeriod())
+          )
+          && loanPolicy.getRenewalsPolicy().getAlternateFixedDueDateScheduleId() != null
+       )) {
+      String message =
```
craigmcnally, line 154: Consider using a stringBuilder here

src/main/java/org/folio/rest/impl/LoanPoliciesAPI.java
```diff
@@ -110,7 +113,103 @@ public void putLoanPolicyStorageLoanPoliciesByLoanPolicyId(
...
+        loanPolicy.getLoanable() + " and profile is of type fixed";
+      respond422("fixedDueDateScheduleId",
+          loanPolicy.getLoansPolicy().getFixedDueDateScheduleId(),
+          message, asyncResultHandler);
+      return;
+    }
+
+    // alternate renewal loan period
+    if (isFixed
+        && is(loanPolicy.getRenewable())
```
craigmcnally, line 188: The formatting of these large conditionals is inconsistent, please pick a format and use it throughout

julianladisch: The formatting is exactly the formatting of the *.drl files that are to be deleted.

julianladisch: The formatting is kept to make code review easy.

marcjohnson-kint: Personally, from my reviewing perspective, I'm not comparing this code to the deleted drools files, I'm considering this as new code. I don't think the formatting needs to be retained.

julianladisch:
Unblocking this pull request because master with #250 has been merged.

PR #253, opened 2020-07-24
CIRCSTORE-223: Add loan.status.name b-tree index.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-223 - Create b-tree index for loan status.name

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -101,6 +101,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "status.name",
+          "tOps": "ADD",
+          "caseSensitive": false,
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -101,6 +101,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "status.name",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
```
marcjohnson-kint, line 109: Has it been confirmed that this setting will align with a general CQL query for this property. There is a mixture of definitions for the other indexes.

bohdan-suprun: I have copied format from other `status.name` indexes.

PR #255, opened 2020-08-03
CIRCSTORE-230: Add loan.itemStatus b-tree index.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-230 - Add b-tree index for loan.itemStatus

### DIFF
--- a/src/test/java/org/folio/rest/api/LoansApiTest.java
+++ b/src/test/java/org/folio/rest/api/LoansApiTest.java
@@ -1329,9 +1329,9 @@ public class LoansApiTest extends ApiTests {
   @Test
   public void canSearchByLoanStatus() throws Exception {
     final IndividualResource openLoan = loansClient.create(
-      new LoanRequestBuilder().open().checkedOut());
+      new LoanRequestBuilder().checkedOut());
 
-    loansClient.create(new LoanRequestBuilder().closed().checkedOut());
+    loansClient.create(new LoanRequestBuilder().checkedIn());
 
     final List<String> openLoans = loansClient.getMany("status.name==Open")
       .getRecords().stream()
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -1329,9 +1329,9 @@ public void cannotProvideAdditionalPropertiesInLoanStatus()
   @Test
   public void canSearchByLoanStatus() throws Exception {
     final IndividualResource openLoan = loansClient.create(
-      new LoanRequestBuilder().open().checkedOut());
+      new LoanRequestBuilder().checkedOut());
 
-    loansClient.create(new LoanRequestBuilder().closed().checkedOut());
+    loansClient.create(new LoanRequestBuilder().checkedIn());
```
marcjohnson-kint, line 1334: Given this search is about the loan status, rather than the item status, why has the explicit specification of a loan status been removed?

src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -1342,6 +1342,22 @@ public void canSearchByLoanStatus() throws Exception {
...
   }
 
+  @Test
+  public void canSearchByLoanItemStatus() throws Exception {
+    final IndividualResource agedToLostLoan = loansClient.create(
+      new LoanRequestBuilder().agedToLost());
+
+    loansClient.create(new LoanRequestBuilder().lostAndPaid());
+
+    final List<String> agedToLostLoans = loansClient.getMany("itemStatus==Aged to lost")
```
marcjohnson-kint, line 1352: Why does the builder define a loan that is `lost and paid` and yet the query tries to find a loan where the item is `aged to lost`?

Does `lostAndPaid` specify more than one aspect of the loan?

src/test/java/org/folio/rest/api/LoansApiTest.java
bohdan-suprun: The status has been encapsulated in the checkIn/out methods of loan builder.

src/test/java/org/folio/rest/api/LoansApiTest.java
bohdan-suprun: there is one more loan declaration above the lostAndPaidLoan.

PR #269, opened 2020-11-12
CIRCSTORE-247 - Request log actions: Pickup expired, Expired

khandramai:
[CIRCSTORE-247](https://issues.folio.org/browse/CIRCSTORE-247) - Request log actions: Pickup expired, Expired


khandramai:
+ @siarhei-charniak

descriptors/ModuleDescriptor-template.json
```diff
@@ -1,6 +1,28 @@
 {
   "id": "${artifactId}-${version}",
   "name": "Circulation Storage Module",
+  "requires": [
+    {
+      "id": "login",
+      "version": "7.0"
+    },
```
bohdan-suprun, line 8: Is this needed because the request expiration process is not an OKAPI timer event but Vert.x periodic task, so we don't have neither okapi-url nor user-token?

If yes, then we should just convert the expiration process to an OKAPI scheduled job. What do you think?

khandramai: Yes, you are correct. Could you please provide more details on OKAPI scheduled job?

khandramai: Do you mean timer interface https://github.com/folio-org/okapi/blob/master/doc/guide.md#timer-interface?

khandramai: @bohdan-suprun @marcjohnson-kint  I tried to minimize interference with the existing logic. If you guys think that timer interface will be more suitable then I will redo this PR. Let me please know your opinion.

bohdan-suprun: Yes, I've pushed some draft changes for this, see [CIRCSTORE-247-use-okapi-scheduled-jobs-for-rq-expiration branch](https://github.com/folio-org/mod-circulation-storage/compare/CIRCSTORE-247-use-okapi-scheduled-jobs-for-rq-expiration?expand=1)

marcjohnson-kint: I think it would be preferable to use the okapi timer interface. The built in RAML Module Builder timed execution is pretty much deprecated and contributes to some operational challenges.

I don't think it is appropriate to impersonate the `pub-sub` user during these processes (it has sometimes been an unfortunate side effect of using pub-sub in other processes due the how message delivery works). I understand it is convenient because that user already exists. It creates a subtle coupling between this module and mod-pubsub.

If this code is switched to use the okapi timer interface, this process will effectively be anonymous, meaning there will not be a user associated with any of these log entries.

In general, I think that storage modules should try to avoid dependencies wherever possible. (this is based upon my understanding of the original intent of storage modules, that I'm trying to verify)

My personal preference would've been for these activities to be moved to mod-circulation and then extends to publish messages for the circulation log. I imagine we aren't in a position to do that at this point.

Mikhail sought a special case for a dependency on `Kafka` from `mod-inventory-storage` yesterday and `mod-source-record-storage` already has dependencies on `pub-sub` so the dependencies are creeping in anyway.

khandramai: OK, guys. I start on refactoring. Thank you!

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -44,11 +45,19 @@ public void postTenant(TenantAttributes ta, Map<String, String> headers,
               .respond500WithTextPlain(res1.cause().getLocalizedMessage())));
             return;
           }
-          hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
-            .respond201WithApplicationJson("")));
+          PubSubRegistrationService.registerModule(headers, vertx)
+          .handle((aBoolean, throwable) -> {
```
bohdan-suprun, line 49: We can use `whenComplete` here so we won't need `return null` at the end.
```suggestion
          .whenComplete((aBoolean, throwable) -> {
```

descriptors/ModuleDescriptor-template.json
bohdan-suprun, line 8: Could this dependency be removed now, since we don't log in anymore?

siarhei-charniak: Yes, sure

src/main/java/org/folio/service/PubSubRegistrationService.java
```diff
@@ -0,0 +1,71 @@
...
+
+    try {
+      for (EventType eventType : EventType.values()) {
+        CompletableFuture<Boolean> future = new CompletableFuture<>();
+        client.deletePubsubEventTypesPublishersByEventTypeName(eventType.name(), PubSubClientUtils.constructModuleName(), ar -> {
+          if (ar.statusCode() == HTTP_NO_CONTENT.toInt()) {
+            future.complete(true);
+          } else {
+            ModulePubSubUnregisteringException exception = new ModulePubSubUnregisteringException(
+                String.format("Module's publisher for " + "event type %s was not unregistered from PubSub. HTTP status: %s",
```
bohdan-suprun, line 55:
```suggestion
                String.format("Module's publisher for event type %s was not unregistered from PubSub. HTTP status: %s",
```

src/main/java/org/folio/support/ExpirationTool.java
```diff
@@ -212,14 +223,18 @@ private static Request changeRequestStatus(Request request) {
...
 
     Future<Void> future = succeededFuture();
     Set<String> closedRequestsItemIds = new HashSet<>();
 
     for (Request request : requests) {
+      JsonObject pair = new JsonObject();
+      pair.put(ORIGINAL.value(), JsonObject.mapFrom(request));
       closedRequestsItemIds.add(request.getItemId());
       Request updatedRequest = changeRequestStatus(request).withPosition(null);
+      pair.put(UPDATED.value(), JsonObject.mapFrom(updatedRequest));
```
bohdan-suprun, line 236: Do you need all the properties in request or just certain fields? Will it be handled properly if a new property is added?

siarhei-charniak: It is better to have all the properties for the possible future changes, new properties will not affect the result

PR #271, opened 2020-11-17
CIRCSTORE-257: Loan policy: Allow recalls to extend overdue loans

kaladay:
## Purpose
Provide the storage for the Boolean and Period representing the Alternate Recall Return and Alternate Recall Return Interval.

Resolves: https://issues.folio.org/browse/CIRCSTORE-257

## Approach
Update the loan-policy raml and its respective version.
Update the appropriate examples.
Update the relevant tests.
I also noticed that the example loan policy could be more thorough of an example.

### DIFF
--- a/ramls/loan-policy.json
+++ b/ramls/loan-policy.json
@@ -109,6 +109,15 @@
               "type": "object",
               "$ref": "period.json",
               "description": "Recall return interval"
+            },
+            "alternateRecallReturn": {
+              "type": "boolean",
+              "description": "Allow recall return interval for overdue loans"
+            },
+            "alternateRecallReturnInterval": {
+              "type": "object",
+              "$ref": "period.json",
+              "description": "Alternate recall return interval for overdue loans"
             }
           }
         },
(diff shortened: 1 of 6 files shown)


ramls/loan-policy.json
```diff
@@ -109,6 +109,15 @@
...
               "description": "Recall return interval"
+            },
+            "alternateRecallReturn": {
+              "type": "boolean",
+              "description": "Allow recall return interval for overdue loans"
+            },
+            "alternateRecallReturnInterval": {
+              "type": "object",
+              "$ref": "period.json",
+              "description": "Alternate recall return interval for overdue loans"
```
bohdan-suprun, line 120: I'd personally keep only the period property and if it exists - it means that alternate recall return interval is true, otherwise - false. 

If we still need two properties, then we probably should add a validation so that `alternateRecallReturnInterval` is specified if the flag is true and is not specified if the flag is false.

What do you think?

kaladay: Makes sense. I suppose I was being too literal in following the mockups.

marcjohnson-kint: There is a part of me that likes the combination of the two properties, it makes the choice as to whether an alternate was chosen explicit and (if the combined validation is provided) makes accidental configuration less likely.

bohdan-suprun (approved):

PR #275, opened 2020-11-19
CIRCSTORE-260: Anonymize loan history entries as well as loan.

bohdan-suprun:
https://issues.folio.org/browse/CIRCSTORE-260 - Anonymize all versions of loans when anonymizing

### Changes
* Fix the SQL query so that the loan history is also anonymized.

### DIFF
--- a/src/main/java/org/folio/rest/impl/AnonymizeStorageLoansAPI.java
+++ b/src/main/java/org/folio/rest/impl/AnonymizeStorageLoansAPI.java
@@ -118,14 +118,14 @@ public class AnonymizeStorageLoansAPI implements AnonymizeStorageLoans {
     // Only anonymize the history for loans that are currently closed
     // meaning that we need to refer to loans in this query
     final String AnonymizeStorageLoansActionHistorySql = String.format(
-      new StringBuilder().append("UPDATE %s_%s.%s l")
+      new StringBuilder().append("UPDATE %s_%s.%s al")
         .append(" SET jsonb = jsonb #- '{loan,userId}'")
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/impl/AnonymizeStorageLoansAPI.java
```diff
@@ -118,14 +118,14 @@ private String createAnonymizationSQL(@NotNull Collection<String> loanIdList,
...
         .append(" SET jsonb = jsonb #- '{loan,userId}'")
         .append(" WHERE jsonb->'loan'->>'id' IN")
         .append(" (SELECT l.jsonb->>'id' FROM %s_%s.loan l")
         .append(" WHERE l.id in ")
         .append(loanIds)
         .append(" AND l.jsonb->'status'->>'name' = 'Closed')")
-        .append(" AND  l.jsonb->'userId' is NOT NULL")
+        .append(" AND  al.jsonb->'loan'->>'userId' is NOT NULL")
         .toString(),
       tenantId, MODULE_NAME, LOAN_HISTORY_TABLE, tenantId, MODULE_NAME);
```
marcjohnson-kint, line 130: Does this mean the logic for anonymising loan history was already present (it just wasn't working)?

marcjohnson-kint (approved):

src/main/java/org/folio/rest/impl/AnonymizeStorageLoansAPI.java
bohdan-suprun: Yes, correct

PR #282, opened 2021-02-09
added patronId b-tree index on patron_action_session table

mtraneis:
Add two b-tree indexes on the fields patronId and actionType that are missing when processing patron notices.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -318,7 +318,21 @@
     {
       "tableName": "patron_action_session",
       "withMetadata": true,
-      "withAuditing": false
+      "withAuditing": false,
+      "index": [
+        {
+          "fieldName": "patronId",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        },
+        {
+          "fieldName": "actionType",
+          "tOps": "ADD",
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -318,7 +318,21 @@
...
+      "withAuditing": false,
+      "index": [
+        {
+          "fieldName": "patronId",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        },
+        {
+          "fieldName": "actionType",
```
julianladisch, line 330: This index doesn't make sense.
actionType is a required field, and the allowed values are "Check-out" and "Check-in". This will match 51% and 49% of all records. Are there any use cases where this can be useful AND the index will be used?

mtraneis: @julianladisch The logs show warnings of missing this index. Certainly the number of warnings are not as numerous as the patronId one.  You don't think that matching 50% of the time of all records is useful?

mtraneis: @julianladisch I removed it. There are only a few actionType options so having an index on it is essentially useless.

julianladisch (approved):
Looks good, thanks!

PR #283, opened 2021-02-17
CIRCSTORE-266: Upgrade to RAML Module Builder 32.x

oleksandr-yatsenko1:
## Purpose
[CIRCSTORE-266](https://issues.folio.org/browse/CIRCSTORE-266)
As part of 2021 R1 it is necessary to upgrade all modules based upon RAML Module Builder to the official version, 32.x

## Approach
Needs to update the module to RAML Module Builder versions of 32.x. As part of this update these points needs to be considered:
- changes to the Tenant API
- upgrading to Vert.x 4.x

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [ ] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [x] There are no breaking changes in this PR.
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation-storage</artifactId>
   <groupId>org.folio</groupId>
-  <version>12.2.0-SNAPSHOT</version>
+  <version>12.2.1-SNAPSHOT</version>
```
marcjohnson-kint, line 5: Why has the module version been changed?

pom.xml
```diff
@@ -110,13 +110,19 @@
...
       <artifactId>rest-assured</artifactId>
-      <version>${rest-assured.version}</version>
+      <version>4.3.1</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.sun.xml.bind</groupId>
+          <artifactId>jaxb-osgi</artifactId>
+        </exclusion>
+      </exclusions>
```
marcjohnson-kint, line 120: What is the purpose of this exclusion? Maybe a comment could be added to explain to future readers why this needs to be excluded.

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -326,7 +321,7 @@ private static void startVerticle(DeploymentOptions options)
     deploymentComplete.get(30, TimeUnit.SECONDS);
   }
 
-  private static void prepareTenant(String tenantId) {
+/*  private static void prepareTenant(String tenantId) {
```
marcjohnson-kint, line 324: Why is this code commented out rather than removed?

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -396,5 +391,65 @@ private static void removeTenant(String tenantId) {
...
+    jo.put("parameters", ar);
+    if (moduleFrom != null) {
+      jo.put("module_from", moduleFrom);
+    }
+    jo.put("module_to", moduleTo);
+    tenantOp(tenantId, jo);
+  }
+
+  private static void removeTenant(String tenantId) {
+
```
marcjohnson-kint, line 417: For methods where the signature does not span multiple lines, I don't think the whitespace gaps improves readability, please remove it.

src/test/resources/log4j2-test.properties
```diff
@@ -0,0 +1,19 @@
...
+appenders = console
+
+appender.console.type = Console
+appender.console.name = STDOUT
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{HH:mm:ss} [$${FolioLoggingContext:requestid}] [$${FolioLoggingContext:tenantid}] [$${FolioLoggingContext:userid}] [$${FolioLoggingContext:moduleid}] %-5p %-20.20C{1} %m%n
+
+rootLogger.level = info
+rootLogger.appenderRefs = info
+rootLogger.appenderRef.stdout.ref = STDOUT
```
marcjohnson-kint, line 19: I thought the guidance had been updated that this file is no longer needed for an upgrade?

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
```diff
@@ -118,7 +118,7 @@ public void putScheduledNoticeStorageScheduledNoticesByScheduledNoticeId(String
   private Future<Void> executeSql(PostgresClient pgClient, String sql) {
 
     Promise<RowSet<Row>> promise = Promise.promise();
-    pgClient.execute(sql, promise.future());
+    pgClient.execute(sql, promise);
     return promise.future().map(ur -> null);
```
julianladisch, line 122: The body of executeSql can be replaced by

    return pgClient.execute(sql).mapEmpty();

Hoiwever, as the return value of executeSql is not used the method executeSql should be deleted and deleteScheduledNoticeStorageScheduledNotices should directly call

    pgClient.execute(sql)

pom.xml
```diff
@@ -110,13 +110,19 @@
...
     </dependency>
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>mod-pubsub-client</artifactId>
-      <version>1.3.1</version>
+      <version>1.3.3</version>
```
marcjohnson-kint, line 125: I thought this work was blocked by an upgrade of the `mod-pubsub-client` that supports vert.x 4.x. This version does not seem to do that. Was I mistaken?

pom.xml
oleksandr-yatsenko1: As in case with mod-patron, it's wrong updating of snapshot number, because doesn't make sense. Will fix it, thanks.

pom.xml
oleksandr-yatsenko1: Good point, this was excluded because of version conflict with transitive dependency carried by the Spring's artifact, as I remember. I'll check it again to make sure this is needed for now.

src/test/java/org/folio/rest/api/StorageTestSuite.java
oleksandr-yatsenko1: Yes, this needs to be removed, thanks.

src/test/java/org/folio/rest/api/StorageTestSuite.java
oleksandr-yatsenko1: Done

src/test/resources/log4j2-test.properties
oleksandr-yatsenko1: Yes, will remove it. Thank you Marc.

pom.xml
oleksandr-yatsenko1: I suspected that version currently set for pubsub client, which is 1.3.3, doesn't contains recently merged changes. But I just wondering can be 1.4.0-SNAPSHOT set for now or only remains is to wait until new release of module will be emitted? I have concerns regarding timing of dependent module release, except the case when we asking folijet team to force module release to resolve dependency issues. What do you think, what approach is better in this situation? By the way, tests were passed because it looks I had disabled test case uses pubsub client, will revert it back to test suite.

marcjohnson-kint: > But I just wondering can be 1.4.0-SNAPSHOT set for now or only remains is to wait until new release of module will be emitted? I have concerns regarding timing of dependent module release, except the case when we asking folijet team to force module release to resolve dependency issues. What do you think, what approach is better in this situation? 

Personally, I would prefer not to use `SNAPSHOT` dependencies because they create non-repeatable builds. 

It was my understanding that Folijet had agreed to release it early to mitigate this challenge.

Please do try the `SNAPSHOT` dependency to check if it works. I don't think this pull request should be merged until there is a non-snapshot artefact to depend upon.

> By the way, tests were passed because it looks I had disabled test case uses pubsub client, will revert it back to test suite.

Thank you for re-including those tests

pom.xml
oleksandr-yatsenko1: I have checked and it looks like currently there is no dependencies conflicts, therefore will remove this. Thank you Marc for noticing this.

src/main/java/org/folio/rest/impl/ScheduledNoticesAPI.java
oleksandr-yatsenko1: Good point, thank you Julian.  Will fix this.

pom.xml
oleksandr-yatsenko1: Thank you Marc for sharing your point of view for this question. I just have discussed this point with team responsible for pubsub client too, they planning to prepare release of version 1.4 this sprint. So for now I'll leave SNAPSHOT there and will update dependency once new version be released.

marcjohnson-kint (approved):

pom.xml
oleksandr-yatsenko1: I have updated the version of dependency of pubsub-client to latest released one - 2.0.0.

PR #285, opened 2021-03-11
Test containers and aspectj fix

adamdickmeiss:


src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -128,44 +129,7 @@ public static void before()
...
-
-      default:
-        String message = "No understood database choice made."
-          + "Please set org.folio.circulation.storage.test.config"
-          + "to 'external', 'environment' or 'embedded'";
-
-        log.error(message);
-        assert false;
-    }
+    PostgresClient.setPostgresTester(new PostgresTesterContainer());
```
marcjohnson-kint, line 132: This is much nicer than what I used to need to do, thank you.

pom.xml
```diff
@@ -45,10 +45,16 @@
       <artifactId>joda-time</artifactId>
       <version>2.10.6</version>
     </dependency>
+    <dependency>
```
marcjohnson-kint, line 48: Please could this be moved to be with the other test dependencies.

marcjohnson-kint (commented):
I've tested this with the docker containers and it works well for me.

src/test/java/org/folio/rest/api/StorageTestSuite.java
adamdickmeiss: Thx

pom.xml
adamdickmeiss: Done

marcjohnson-kint:
@adamdickmeiss I haven't had chance to test that running the tests with an external PostgreSQL instance works. Have you tested that?

adamdickmeiss:
> @adamdickmeiss I haven't had chance to test that running the tests with an external PostgreSQL instance works. Have you tested that?

I did. It will *not* at this stage use DB_-variables when the PostgresClient.setPostgresTester is called.

It begs the question what is the right thing to do:

1. consider this a bug and make PostgresClient.setPostgresTester do "nothing" if DB_-env is is used.

2. We don't change anything but require applications to decide on their own.

3. 3rd trigger.

The first option is certainly appealing . The fix in RMB is a one line removal. As a developer I don't like it. When I set these environment varibles it's because I'm doing tests OUTSIDE unit tests.. Since thse DB-things are also used in "production" and all, this could lead to nasty surprises.

Second options is what have had so far.. And we change nothing in RMB. This very PR must have a way to trigger the PostgresClient.setPostgresTester. .. Each developer will cut'n paste from previous project and it will get quite ugly.

Third option is to have another, specifically for people who wish to run their native postgres for testing.

I always prefer the containers over my own postgres. It's like 1 second apart in mod-circulation-storage  from each test a total run of 26 seconds.

marcjohnson-kint:
> 1. consider this a bug and make PostgresClient.setPostgresTester do "nothing" if DB_-env is is used.
> 2. We don't change anything but require applications to decide on their own.
> 3. 3rd trigger.
> 
> The [change] in RMB is a one line removal. As a developer I don't like it. When I set these environment variables it's because I'm doing tests OUTSIDE unit tests.. Since these DB-things are also used in "production" and all, this could lead to nasty surprises.

Why don't you like it? What kind of nasty surprises might come from the tests using the same configuration mechanism as production code?

> Second option is what have had had so far.. And we change nothing in RMB. This very PR must have a way to trigger the PostgresClient.setPostgresTester. .. Each developer will cut'n paste from previous project and it will get quite ugly.

I wrote that code to allow me to run the tests often without a long delay. I didn't intend for it to be copied and pasted around a lot. I agree that causes some challenges (though if it is stable, I'm not sure it really matters). I have tended not to try to push stuff back to the shared libraries.

> I always prefer the containers over my own postgres. It's like 1 second apart in mod-circulation-storage from each test a total run of 26 seconds.

I think that relies on folks always using containers, and particularly a container controlled by the tests themselves.

marcjohnson-kint (approved):
I think we should consider supporting an external PostgreSQL in future. Irrespective of that, I think this is likely better than what we currently have with embedded PostgreSQL.

@K-Felk does this pull request resolve your stability issues?

K-Felk:
> @K-Felk does this pull request resolve your stability issues?

Yes and no-it resolves the issue I was seeing where all tests were failing with messages that the back end database was not operational.  Tests are now failing with a different error message that seems to have nothin to do with the database.  that may well be due to some kind of fault in my own environment-I'm just not sure.  I've detailed what's happening in the slack discussion thread-it might be better to hash that out there.

K-Felk (approved):

adamdickmeiss:
Thx for looking at this.

Let me know if I should re-enable the code to check org.folio.circulation.storage.test.database.

PR #286, opened 2021-03-16
CIRCSTORE-273 include migration for missing staff slips

zburke:
The reference data includes four slips (hold, pick-slip,
request-delivery, and transit) but only two (pick-slip,
request-delivery) were included in migrations. This meant new tenants
would have all four slips available but tenants migrated from previous
releases would only have two.

Refs [CIRCSTORE-273](https://issues.folio.org/browse/CIRCSTORE-273)

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -353,7 +353,7 @@
     {
       "run": "after",
       "snippetPath": "add_staff_slips_hold_transit.sql",
-      "fromModuleVersion": "12.2.0"
+      "fromModuleVersion": "12.3.0"
     }
   ]
 }
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -349,6 +349,11 @@
     {
       "run": "after",
       "snippetPath": "index_dateLostItemShouldBeBilled.sql"
+    },
+    {
+      "run": "after",
+      "snippetPath": "add_staff_slips_hold_transit.sql",
+      "fromModuleVersion": "12.2.0"
     }
   ]
 }
--- a/src/test/java/org/folio/rest/api/migration/StaffSlipsHoldTransitMigrationScriptTest.java
+++ b/src/test/java/org/folio/rest/api/migration/StaffSlipsHoldTransitMigrationScriptTest.java
@@ -1,11 +1,12 @@
 package org.folio.rest.api.migration;
 
-import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
 
-import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.JsonObject;
-import org.apache.commons.lang3.StringUtils;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
 import org.folio.rest.api.StorageTestSuite;
 import org.folio.rest.support.JsonResponse;
 import org.folio.rest.support.ResponseHandler;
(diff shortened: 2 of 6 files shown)


marcjohnson-kint (approved):

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -349,6 +349,11 @@
     {
       "run": "after",
       "snippetPath": "index_dateLostItemShouldBeBilled.sql"
+    },
+    {
+      "run": "after",
+      "snippetPath": "add_staff_slips_hold_transit.sql",
+      "fromModuleVersion": "12.2.0"
```
alexanderkurash, line 356: It looks like we're adding hold and transit slips with a separate script and setting different `fromModuleVersion` for it (`12.2.0` vs `11.1.0`). Is this how it's supposed to work? What should happen if the user upgrades to a version in between `11.1.0` and `12.2.0`? Should we consider just adding these two slips to the existing script?

src/test/java/org/folio/rest/api/migration/StaffSlipsHoldTransitMigrationScriptTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+public class StaffSlipsHoldTransitMigrationScriptTest extends MigrationTestBase {
```
alexanderkurash, line 22: Same question here - if we include these two slips in the existing script, we won't need to create another test class. Though, if we do need to to have a separate script, we can move most of the code of this class to `MigrationTestBase` because it duplicates `StaffSlipsMigrationScriptTest` code.

src/test/java/org/folio/rest/api/migration/StaffSlipsHoldTransitMigrationScriptTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    assertThat(staffSlip.getString("id"), is(expectedId));
+    assertThat(staffSlip.getString("name"), is(expectedName));
+    assertThat(staffSlip.getBoolean("active"), is(true));
+    assertThat(staffSlip.getString("template"), is(TEMPLATE));
+  }
+
+  private URL staffSlipsStorageUrl(String subPath) throws MalformedURLException {
+    return StorageTestSuite.storageUrl("/staff-slips-storage/staff-slips" + subPath);
+  }
```
alexanderkurash, line 71: For example, these methods are common between `StaffSlipsMigrationScriptTest` and `StaffSlipsHoldTransitMigrationScriptTest`, I would consider moving them to a single place.

src/main/resources/templates/db_scripts/schema.json
zburke: The details are now hazy, but IIRC, I asked somebody about just augmenting the existing script and was told a separate script would be better because updating an existing script would effectively prevent customers who have already run that from accessing the migration. But your comment about `fromModuleVersion` absolutely makes sense; maybe the right thing is to use multiple scripts but with the same value there?

marcjohnson-kint: @alexanderkurash 

bq. Should we consider just adding these two slips to the existing script?

As I understand it, RAML Module Builder uses the association between each SQL script and a version to determine whether to execute that SQL. I don't understand the logic used completely, I believe that amending an existing script for a version that has already been deployed will mean those changes are not applied.

> What should happen if the user upgrades to a version in between 11.1.0 and 12.2.0?

I'm not sure I understand this question. We are adding the new slip definitions in version 12.2.0 as that is the next intended release. Are you asking if we should be also back porting this change to previous releases?

marcjohnson-kint: Also, `12.2.0` has already been released, which I believe means that the `fromModuleVersion` for this script needs to be `12.3.0` for this to be executed on existing systems.

alexanderkurash: Ok, second script makes sense to me now.

> I'm not sure I understand this question. We are adding the new slip definitions in version 12.2.0 as that is the next intended release. Are you asking if we should be also back porting this change to previous releases?

Yes, this is what I meant. I believe we've added those slips a long time ago (in 2019), but didn't add them to the migration script. If we set `fromModuleVersion` to the latest version, existing systems would need to be updated to this version in order to get these slips. I don't know, maybe this is a good thing if no one is planning to update to older versions.

> Also, 12.2.0 has already been released, which I believe means that the fromModuleVersion for this script needs to be 12.3.0 for this to be executed on existing systems.

Should it be 12.2.1, considering that CIRCSTORE-273 is marked as R1 bugfix?

marcjohnson-kint: > I don't know, maybe this is a good thing if no one is planning to update to older versions.

If folks want an update to the older distributions of FOLIO, then I think they will need to ask for this change to be back ported.

> Should it be 12.2.1, considering that CIRCSTORE-273 is marked as R1 bugfix?

I don't have an authoritative answer for you. Maybe the RAML Module Builder folks can advise on that.

My personal preference is for the script to be marked as _from_ `12.3.0` on the mainline and `12.2.1` on the patch release branch. This is due to `12.2.1` not being a legitimate version for the mainline and so this configuration could be confusing (and I personally find this area really confusing). This does mean that the script needs to be repeatable, as folks can upgrade from `12.2.1` to `12.3.0`, and it does mean the back port needs to change the code.

alexanderkurash (approved):

PR #287, opened 2021-03-19
CIRCSTORE-186 Use normalization for request.item.identifiers

K-Felk:
The code adding the index is, I think, very straightforward
(thank you to Bohdan for helping me get it working).

The test may require a bit more explanation.  I originally had this
written as a series of separate tests, but I was unable to get the setup
code to execute that way-beforeAll requires a static function, and 
I needed a lot of non-static methods to do the setup, so I consolidated 
all the tests into one test and use @Before.  I originally tried situating 
inside some of the existing tests, but found I needed a lot of code 
specifically written to support the search and retrieval, plus some very 
test-specific setup (records with very specific item isbns) so I ended 
up putting it in its own file.

refs: https://issues.folio.org/browse/CIRCSTORE-186


NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 12.3.0 IN-PROGRESS
+
+* implements normalization on ISBN index (CIRCSTORE-186)
```
marcjohnson-kint, line 3: How would a client use this, is it an index to be used in a CQL query?

reference-data/identifier-types/isbn.json
```diff
@@ -0,0 +1,5 @@
+{
+    "id":"8261054f-be78-422d-bd51-4ed9f33c3422",
+    "name":"isbn",
+    "source": "folio"
+}
```
marcjohnson-kint, line 5: As far as I know, circulation-storage does not provide a set of `identifier types`. What is this used for?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -115,6 +115,11 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "identifiers",
+          "arraySubfield" : "value",
+          "arrayModifiers": ["identifierTypeId"]
         }
```
marcjohnson-kint, line 123: This index appears to have been added to the top-level `identifiers` property of loans. 

My understanding is that this work is intended to facilitate searching requests by isbn.

Please can you help me understand the intent of this index?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -167,6 +172,18 @@
           "removeAccents": true
         }
       ],
+      "fullTextIndex": [
+      	{
+          "fieldName": "identifiers",
+          "arraySubfield" : "value",
+          "arrayModifiers": ["identifierTypeId"]
+        },
```
marcjohnson-kint, line 180: As I understand it, the `identifiers` array is part of the `item` object for a request. This index appears to refer to a top level property. Have I missed something?

NEWS.md
K-Felk: Yup.  I'll replace this with the more verbose description in the jira issue-that should be clearer.

reference-data/identifier-types/isbn.json
K-Felk: The model I was following for this work had a similar file-I assumed it was necessary for associating this specific UUID with isbn identifiers, so that they could be retrieved by that UUID later (for example in tests, which pull the identifiers based on the UUID provided here).  When I examined the isbn identifiers in the tests, they had this UUID for isbn identifier type, which seemed to confirm that theory.

src/main/resources/templates/db_scripts/schema.json
K-Felk: My apologies-at an early stage of this work, I had this incorrectly positioned in the loans table-I later realized my mistake and put it in the correct place (if you scroll down, you'll see there's identical notation in the request table) I must have forgotten to remove the earlier entry.  I'll fix.

.factorypath
```diff
@@ -0,0 +1,153 @@
+<factorypath>
```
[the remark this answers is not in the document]
K-Felk: Ugh, this thing.  It's a maven artefact.  I was doing a lot of branch-juggling with this work because of the problems I was having with tests crashing-I must have missed it.  I'll remove.

src/main/resources/templates/db_scripts/schema.json
K-Felk: It is, yes.  I was unsure if I needed to put this in some kind of hierarchical notation, or if the schema would simply locate the field anywhere in the hierarchy based on the provided name.  When I tried it this way, it seemed to work, so I assumed the latter.

marcjohnson-kint: > It is, yes. I was unsure if I needed to put this in some kind of hierarchical notation, or if the schema would simply locate the field anywhere in the hierarchy based on the provided name

I think there is some hierarchical notation needed. I don't think the schema tries to match this against any property.

> When I tried it this way, it seemed to work, so I assumed the latter.

How did you know it worked?

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
```diff
@@ -0,0 +1,173 @@
...
+public class IsbnNormalizationTest extends ApiTests {
+  private static final String REQUEST_STORAGE_URL = "/request-storage/requests";
+     // Interesting Times has two ISBNs: 0552167541, 978-0-552-16754-3
+
+  @Before
+  public void setUp() throws InterruptedException, ExecutionException, TimeoutException, MalformedURLException {
+	  createRequests();
+	  Awaitility.await()
+      .atMost(10, TimeUnit.SECONDS)
+      .until(requestsLoaded(), is(true));
```
marcjohnson-kint, line 44: Is the intent of this that the wait until all of the requests have been created?

If so, I don't think this is necessary as the entity creation process only finishes once the request has finished.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    find("isbn = 9780552167543",      "Interesting Times");
+
+    find("isbn = 9780*", "Interesting Times", "Temeraire");  
+    
+    find("isbn = 9-7-8-055-2167-543", "Interesting Times");
+
+    find("isbn = 552-16754-3");
+
+    find("isbn = 552*");
```
marcjohnson-kint, line 62: Do each of these examples demonstrate the same behaviour, specific separate behaviour or are there some sub-categories e.g.
* should find single request by normalized ISBN 
* should find multiple requests by normalized ISBN 
* should find no requests for normalized ISBN that does not exist

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
K-Felk: I assumed so as well at first, but that's not what was actually happening.  Tests would fail, indicating there was no record matching the search parameters.  When I dumped a search for all ISBNS to screen, what I would see was that there were only one or two requests loaded.  So I added the wait and check code-and lo and behold, I started seeing all the records in the body of the test.

src/main/resources/templates/db_scripts/schema.json
K-Felk: Since this code looks like it's connected to generating the index, if it weren't working, I would expect to see: 1) an error message in the test output indicating the field could not be found and 2) the test would fail because the index isn't properly generated, and the identifier field is therefore not searchable.  I'm assuming a hierarchical notation would look like "item.identifiers"?  If it's wrong, and the isbn index is still being generated (as it seems to be), then is that code necessary at all?  I actually tried removing that chunk entirely, and the search test seems to work without it.

marcjohnson-kint: > Since this code looks like it's connected to generating the index, if it weren't working, I would expect to see: 1) an error message in the test output indicating the field could not be found and 2) the test would fail because the index isn't properly generated, and the identifier field is therefore not searchable. 

As I understand it, the CQL index that is being introduced here is the `isbn` index (created below) rather than an index on the identifiers in general. 

I imagine if this index was removed, the tests would still work as before as they are exercising that index rather than this one.

> I'm assuming a hierarchical notation would look like "item.identifiers"? If it's wrong, and the isbn index is still being generated (as it seems to be), then is that code necessary at all? I actually tried removing that chunk entirely, and the search test seems to work without it.

Ah, it seems you've come to the same conclusion as me above.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
marcjohnson-kint: That is rather strange. I've been trying out some experiments at refactoring these tests (I've only really just started) and I've created a [branch](https://github.com/folio-org/mod-circulation-storage/tree/CIRCSTORE-186-refactor-tests). 

Could you give that code a try and let me know if it runs stably for you.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
K-Felk: Each one tests the normalization in a slightly different way: search for an isbn with dashes, search for an isbn with truncation, search for a second isbn in an item with two, etc.  The original test code I was using as a model had these all separated out, with method names that indicated exactly what searching behavior was being tested ("canSearchForIsbnsWithDashes," for example).  But as I mentioned in the notes above, I could not get the setup code to run properly in that configuration.  @beforeClass/@beforeAll would silently fail, leaving no records loaded-I *think* this is because those annotations require static methods only, and the setup relies on a whole web of methods that are non-static, many of which are in external files (like RequestItemSummary).  I tried looking up some different methods of getting @Before notations to operate in non-static contexts, but they all either crashed the build or seemed to do nothing.  So I finally just put all the searching tests into a single test and used @before to do the setup-this is not ideal from a readability standpoint, but it was the only way I could figure out to get the test to actually run with the necessary setup.

marcjohnson-kint: > But as I mentioned in the notes above, I could not get the setup code to run properly in that configuration.

That's what I'm hoping to be able to do.

Please could you put a comment on each of the find invocations describing the intent like you had for the names when they were individual tests. That could help guide the refactoring I'm trying to do.

Thank you for trying a bunch of approaches. I understand none of them working was frustrating, it can definitely be difficult to set up these tests. The test suite workaround (for slow database setup) I put in place a long time ago has become a hinderance to this in quite a few ways :-/

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
K-Felk: It seems to.  Should I merge it in?

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
K-Felk: Thank you for being patient with me.  I've annotated the individual tests in the latest commit, and I removed the code from the schema concerning the identifiers index.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -167,6 +167,13 @@
           "removeAccents": true
         }
       ],
+      "fullTextIndex": [
+        {
+          "fieldName": "isbn",
```
marcjohnson-kint, line 172: Is this the same name as the index has for items in inventory?

K-Felk: You mean in mod-inventory-storage?  If so, yes, the isbn field there is named exactly the same.  Should they be different?

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
marcjohnson-kint: > Should I merge it in?

You are welcome to merge it in

reference-data/identifier-types/isbn.json
marcjohnson-kint: > I assumed it was necessary for associating this specific UUID with isbn identifiers

I don't think it is needed at all.

> for example in tests, which pull the identifiers based on the UUID provided here

How is that done? As far as I can tell, there is no API for identifier types in mod-circulation-storage and no code to load this record.

> When I examined the isbn identifiers in the tests, they had this UUID for isbn identifier type, which seemed to confirm that theory.

I imagine that is because it is hardcoded in the test records.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
marcjohnson-kint: > Thank you for being patient with me. 

Thank you for being receptive to feedback and for trying to navigate your way through some of the poor code in FOLIO.

> I've annotated the individual tests in the latest commit, and I removed the code from the schema concerning the identifiers index.

Thank you

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: > Should they be different?

They could be different. This index is in reference to requests. Do people say, `let's search for requests by ISBN`?

K-Felk:
I removed the isbn identifier file and merged your test refactor into this PR.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Yes it is.

What I'm trying to suggest is that the `ISBN` index for `items` is used for to `search for items with ISBN X`.

Is that the same for requests? Are we trying to `search for requests with ISBN X` or `search for requests for items with ISBN x`?

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
```diff
@@ -0,0 +1,166 @@
...
+    find("isbn = 9780552167543",      "Interesting Times");
+    //canSearchForSecondIsbnWithMissingHyphensAndTrunation()
+    find("isbn = 9780*", "Interesting Times", "Temeraire");  
+    //canSearchForSecondIsbnWithAlteredHyphens()
+    find("isbn = 9-7-8-055-2167-543", "Interesting Times");
+    //cannotFindIsbnWithTailString()
+    find("isbn = 552-16754-3");
+    //cannotFindIsbnWithInnerStringAndTruncation()
+    find("isbn = 552*");
+  }
```
marcjohnson-kint, line 62: Now that the test request creation runs consistently during these tests, please can these tests be split up into separate examples again.

During each test, please only create the items that are needed to demonstrate that example.

src/main/resources/templates/db_scripts/schema.json
K-Felk: The latter?  (search for requests for items with ISBN x)  So are you suggesting we should rename the index to something like `RequestItemIsbn`?

K-Felk:
I refactored the index test into separate tests, each one loads only titles it needs for that test.  I did always include at least one item request designed not to be retrieved-it seems to me that making sure the test does not retrieve items it shouldn't retrieve is as important as checking that it does retrieve items that it's supposed to retrieve.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: > Are you suggesting we should rename the index to something like `RequestItemIsbn`?

My suggestion would be something like `itemIsbn` given that the index is used in the context of the set of requests, I think the `request` part is redundant

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    find("isbn = 0-552-16754-1",      "Interesting Times");
+  }
+
+  @SneakyThrows
+  @Test
+  public void canSearchForFirstIsbnWithAdditionalHyphenAndTruncation() {
+    createRequests("Interesting Times", "Uprooted");
+
+    find("isbn = 05-5*",              "Interesting Times");
```
marcjohnson-kint, line 58: What is the intent behind the horizontal alignment of the second parameter across tests?

marcjohnson-kint:
> I did always include at least one item request designed not to be retrieved-it seems to me that making sure the test does not retrieve items it shouldn't retrieve is as important as checking that it does retrieve items that it's supposed to retrieve.

Thank you, I agree with that. Only having the intended matches in the set means that it's not possible to know if the search is working.

src/main/resources/templates/db_scripts/schema.json
K-Felk: Okay, I've changed the indexname in the schema and the tests.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
K-Felk: The way I originally wrote this, the tests were all the same line length, and close together, so that the titles and searches all lined up across tests.  It's been fiddled with so much that formatting no longer works.  I'll remove in the next commit.

src/test/java/org/folio/rest/api/IsbnNormalizationTest.java
```diff
@@ -0,0 +1,194 @@
...
+      String url = requestStorageUrl("").toString() + "?query=" + urlEncode(cql);
+
+      client.get(url, TENANT_ID, responseHandler(searchCompleted));
+      Response searchResponse = searchCompleted.get(5, SECONDS);
+
+      assertThat(searchResponse.getStatusCode(), is(200));
+      return searchResponse.getJson();
+    } catch (Exception e) {
+      return new JsonObject("");
+    }
```
marcjohnson-kint, line 142: What is the intent for not propagating exceptions when searching for requests?

Is it to avoid needing to define all of the exceptions that could occur in every client of this code?

K-Felk: The model code I was following had the request written like this-I assumed it was because an error at this stage would indicate something wrong with the testing environment, such as a problem with the database set-up, in which case, all tests would be failing with errors.

marcjohnson-kint: > The model code I was following had the request written like this

Fair enough. It can be difficult to know if that code is an exemplar of a positive pattern or not.

> I assumed it was because an error at this stage would indicate something wrong with the testing environment, such as a problem with the database set-up, in which case, all tests would be failing with errors.

That could be the case. Under those circumstances, I would want the tests to fail with an error (that would hopefully reveal that situation) rather than an assertion mismatch (that I would then need to figure out the cause for) which is what I think this code would do.

K-Felk: Okay, I'll rewrite this to push up errors to the top level.

marcjohnson-kint (approved):

PR #288, opened 2021-03-25
CIRCSTORE-265 - Pickup expired not showing in Circulation log

siarhei-charniak:
[CIRCSTORE-265](https://issues.folio.org/browse/CIRCSTORE-265) - Pickup expired not showing in Circulation log

* Restored Pubsub registration

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -3,11 +3,15 @@ package org.folio.rest.impl;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import java.util.Map;
+
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


khandramai:
@oleksandr-yatsenko1 @marc-johnson ,
PubSub publisher registration was deleted in scope of https://github.com/folio-org/mod-circulation-storage/pull/283 (migration to RMB32). As a result we have an issue https://issues.folio.org/browse/CIRCSTORE-265. We returned back this logic. Could you please review?

marcjohnson-kint (commented):
It seems to me that this was a mistake rather than intentional. Putting it back seems reasonable to me.

My biggest question would be, why didn't any tests fail when the registration was removed?

khandramai:
@marcjohnson-kint, we don't cover this logic (PubSub publisher registration) by unit tests. According to contract we delegate PubSub to create all registration actions. What we only need - send request to PubSub for publisher registration. Ideally, we can catch such issue if we have integration tests for circulation, but as I know they aren't implemented.

khandramai (approved):

alekGbuz (approved):

PR #293, opened 2021-06-09
Upgrade to RMB 33.0.0 CIRCSTORE-280

marcjohnson-kint:
* Upgrades to RMB 33.0.0 and Vert.x 4.1.0.CR1
* Upgrades the pub-sub client as it needs to use the same version of RMB (2.x includes breaking changes, so code needed to be changed)
* The `unregisterModule` method is unused so has been removed
* Removing the `embed_postgres` command line parameter makes this a breaking compatibility change

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
-## 12.3.0 IN-PROGRESS
+## 13.0.0 IN-PROGRESS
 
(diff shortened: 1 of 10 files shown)


K-Felk (approved):
I looked through this and didn't see anything that looked amiss.  I think you can merge.

PR #298, opened 2021-08-16
CIRCSTORE-287: Add @Validate to all API implementations

julianladisch:
Otherwise RMB doesn't validate the input against the API spec.


adamdickmeiss (approved):
Looks good

PR #340, opened 2022-06-02
CIRCSTORE-342 Expire requests for only the tenant the job is being called for

Mykyta-Varenyk:
The expiration job is being called for all of the tenants, and so when the timer is being fired for okapi for two tenants simultaneously two update transactions are executed at the same time and unrepeatable read phenomena occurs. This can be fixed by running the job just for a specific tenant.
Resolves: https://issues.folio.org/browse/CIRCSTORE-342.

### DIFF
--- a/src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
+++ b/src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
@@ -47,6 +47,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import io.vertx.core.json.JsonObject;
+import static org.hamcrest.core.IsNull.*;
 
 public class RequestExpirationApiTest extends ApiTests {
 
@@ -834,7 +835,7 @@ public class RequestExpirationApiTest extends ApiTests {
 
     JsonObject requestFromDummyTenant = getById(requestStorageUrl(String.format("/" + secondRequestId)), dummyTenant);
     assertThat(requestFromDummyTenant.getString("status"), is(CLOSED_PICKUP_EXPIRED));
-    assertThat(requestFromDummyTenant.getInteger("position"), IsNull.nullValue());
+    assertThat(requestFromDummyTenant.getInteger("position"), nullValue());
   }
 
   @SneakyThrows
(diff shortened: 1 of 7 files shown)


src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -790,16 +791,51 @@ public void canExpireOpenAwaitingWithNoHoldShelfExpirationDate()
     assertThat(response.getInteger("position"), is(1));
   }
 
-  private void expireRequests() throws InterruptedException, ExecutionException, TimeoutException, MalformedURLException {
+  @Test
+  @SneakyThrows
+  public void shouldExpireRequestWithAwaitingPickupExpiredFieldWhenExpirationJobIsCalledForTwoTenantsSimultaneously() {
```
OleksandrVidinieiev, line 796: I can see that expiration is called for a single tenant

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -790,16 +792,67 @@ public void canExpireOpenAwaitingWithNoHoldShelfExpirationDate()
...
+        .withRequestExpirationDate(new DateTime(2017, 7, 30, 10, 22, 54, DateTimeZone.UTC))
+        .withHoldShelfExpirationDate(new DateTime(2017, 7, 31, 10, 22, 54, DateTimeZone.UTC))
+        .withItemId(secondItemId)
+        .withPosition(1)
+        .withStatus(OPEN_AWAITING_PICKUP)
+        .create(),
+      requestStorageUrl(), dummyTenant);
+    expireRequestsForTenant(dummyTenant);
+
+    JsonObject requestFromDefaultTenant = getById(requestStorageUrl(String.format("/%s", firstRequestId)));
```
OleksandrVidinieiev, line 830:
```suggestion
    JsonObject requestFromDefaultTenant = getById(requestStorageUrl("/" + firstRequestId));
```

OleksandrVidinieiev (approved):

PR #351, opened 2022-06-30
CIRCSTORE-349 add expirationDate to actualCostRecord

roman-barannyk:
## Purpose 
- Add expirationDate field (date-time), not required.

It'll be used by a timer in mod-circulation that will close loans for items that were aged to lost or declared lost with actual cost in the policy, but Lost Item Fee wasn't charged during a period specified in the policy.

Resolves: [CIRCSTORE-349](https://issues.folio.org/browse/CIRCSTORE-349)

### DIFF
--- a/ramls/actual-cost-record.json
+++ b/ramls/actual-cost-record.json
@@ -118,6 +118,11 @@
       "description": "Fee/fine type name",
       "type": "string"
     },
+    "expirationDate": {
+      "description": "Expiration date and time for items that were lost with actual cost",
+      "type": "string",
+      "format": "date-time"
+    },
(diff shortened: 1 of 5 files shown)


Mykyta-Varenyk (approved):

ramls/actual-cost-record.json
```diff
@@ -118,6 +118,11 @@
       "description": "Fee/fine type name",
       "type": "string"
     },
+    "expirationDate": {
```
roman-barannyk, line 121: increase the interface version

alexanderkurash (approved):

PR #355, opened 2022-09-13
CIRCSTORE-358: Update actualCostRecord schema

OleksandrVidinieiev:
[CIRCSTORE-358](https://issues.folio.org/browse/CIRCSTORE-358)


alexanderkurash (commented):
Should actual cost record example be also changed?

alexanderkurash (approved):

PR #381, opened 2022-12-13
CIRCSTORE-387: RequestExpirationService hangs on publish exception

julianladisch:
In RequestExpirationService.java:

If eventPublisherService.publishLogRecord throws an exception the promise is neither failed nor completed and the request to /scheduled-request-expiration will never get any response.

Approach:

Replace deprecated (and error-prone) PostgresClient.startTx with PostgresClient.withTrans that automatically either commits or rolls back the transaction.


adamdickmeiss (approved):
Good. One step towards getting rid of startTx, which, unfortunately, is used in many places

PR #391, opened 2023-03-15
CIRCSTORE-398 Add effective call number, shelving order and  pickup service point name fields to the request schema

tatsiana-tarhonskaya:
Resolves [CIRCSTORE-398](https://issues.folio.org/browse/CIRCSTORE-398)

## Purpose
Add effective call number, shelving order and service point name fields to the request schema

## Approach
Update schema and add corresponding indexes:

- gin indexes for callNumber and fullCallNumber (for now they're going to be searched as 'starts with' but there're plans to add 'contains' search in future)
- b-tree indexes for shelving order and pickup service point name  - they're going to be used for sorting by these fields


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation-storage</artifactId>
   <groupId>org.folio</groupId>
-  <version>16.1.0-SNAPSHOT</version>
+  <version>16.1.1-SNAPSHOT</version>
```
alexanderkurash, line 5: No need to change module version here, but interface version needs to be changed

alexanderkurash (changes requested):

ramls/request.json
```diff
@@ -226,6 +226,11 @@
       "type": "string",
       "format": "date-time",
       "readonly" : true
+    },
+    "searchIndex": {
+      "description": "Request fields used for search",
+      "type": "object",
+      "$ref": "searchIndex.json"
```
alexanderkurash, line 233: This filename looks too generic. What if we add searchIndex to other objects that would have different indexing fields? We can rename it to requestSearchIndex.json

pom.xml
tatsiana-tarhonskaya: Updated request-storage to 5.1

ramls/request.json
tatsiana-tarhonskaya: thanks, renamed it

alexanderkurash (approved):

PR #401, opened 2023-05-31
CIRCSTORE-408 Update request seach index when service point is updated

alexanderkurash:
Resolves [CIRCSTORE-408](https://issues.folio.org/browse/CIRCSTORE-408)

## Purpose
Search index fields added to request schema in [CIRCSTORE-398](https://issues.folio.org/browse/CIRCSTORE-398) need to be kept up-to-date at all times. In order to achieve this, mod-circulation-storage needs to consume domain events published by mod-inventory-storage every time a service point is updated ([MODINVSTOR-1043](https://issues.folio.org/browse/MODINVSTOR-1043)). When such an event is received, we need to check if service point name was changed, and update its value in the search index of every request where updated service point is specified as pickup service point.

## Approach
An abstract event handler has been created based on the existing item event handler. It now has two implementations - one for item events, another for service point events.


src/test/java/org/folio/rest/api/EventConsumerVerticleTest.java
```diff
@@ -56,8 +58,15 @@ public class EventConsumerVerticleTest extends ApiTests {
   private static final String REQUEST_STORAGE_URL = "/request-storage/requests";
   private static final String KAFKA_SERVER_URL = String.format("%s:%s", host(), port());
   private final static String DEFAULT_PICKUP_SERVICE_POINT_NAME = "Circ Desk 1";
+  private final static String DEFAULT_CALL_NUMBER_PREFIX = "prefix";
+  private final static String DEFAULT_CALL_NUMBER = "callNumber";
+  private final static String DEFAULT_CALL_NUMBER_SUFFIX = "suffix";
+  private final static String DEFAULT_SHELVING_ORDER = "shelvingOrder";
```
OleksandrVidinieiev, line 64: Can we use these in existing tests too?

PR #405, opened 2023-06-13
CIRCSTOR-412: Request cannot be expired properly when TLR settings not set

K-Felk:
Because there are other endpoints that expect to get errors from the settings when TLR settings are not set, I located the fix for this within the expiry code.  If it cannot read the TLR settings from the back end (which is the case when the settings are empty), it creates a dummy set of settings that treat TLR as if it's disabled.  This allows the expiry process to properly expire requests even when TLR settings are not initialized.

### DIFF
--- a/src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
+++ b/src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
@@ -31,20 +31,18 @@ public class RequestExpiryImpl implements ScheduledRequestExpiration {
 
     Vertx vertx = context.owner();
     new ConfigurationClient(vertx, okapiHeaders).getTlrSettings()
-      .onSuccess(tlrSettings -> {
-        doExpiration(tlrSettings, handler, okapiHeaders, vertx);
-      })
-      .onFailure(r -> {
-        TlrSettingsConfiguration emptyConfig = new TlrSettingsConfiguration(false, false, null, null, null);
-        doExpiration(emptyConfig, handler, okapiHeaders, vertx);
-      });
+      .onSuccess(tlrSettings -> doExpiration(tlrSettings, handler, okapiHeaders, vertx))
+      .onFailure(t -> doExpiration(new TlrSettingsConfiguration(false, false, null, null, null), 
+        handler, okapiHeaders, vertx));
   }
 
-  private void doExpiration(TlrSettingsConfiguration tlrSettings, Handler<AsyncResult<Response>> handler, Map<String, String> okapiHeaders, Vertx vertx) {
+  private void doExpiration(TlrSettingsConfiguration tlrSettings,
+    Handler<AsyncResult<Response>> handler, Map<String, String> okapiHeaders, Vertx vertx) {
+    
     createRequestExpirationService(okapiHeaders, vertx, tlrSettings)
-        .doRequestExpiration()
-    .onSuccess(x -> handler.handle(succeededFuture(respond204())))
-    .onFailure(e -> handler.handle(succeededFuture(respond500WithTextPlain(e.getMessage()))));
+      .doRequestExpiration()
+        .onSuccess(x -> handler.handle(succeededFuture(respond204())))
+        .onFailure(e -> handler.handle(succeededFuture(respond500WithTextPlain(e.getMessage()))));
   }
 
   private RequestExpirationService createRequestExpirationService(Map<String, String> okapiHeaders,
--- a/src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
+++ b/src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
@@ -6,6 +6,7 @@ import static org.folio.rest.jaxrs.resource.ScheduledRequestExpiration.Scheduled
 import static org.folio.rest.jaxrs.resource.ScheduledRequestExpiration.ScheduledRequestExpirationResponse.respond500WithTextPlain;
 
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import javax.ws.rs.core.Response;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
```diff
@@ -30,8 +31,18 @@ public void expireRequests(Map<String, String> okapiHeaders,
...
     new ConfigurationClient(vertx, okapiHeaders).getTlrSettings()
-    .compose(tlrSettings -> createRequestExpirationService(okapiHeaders, vertx, tlrSettings)
-        .doRequestExpiration())
+      .onSuccess(tlrSettings -> {
+        doExpiration(tlrSettings, handler, okapiHeaders, vertx);
+      })
+      .onFailure(r -> {
+        TlrSettingsConfiguration emptyConfig = new TlrSettingsConfiguration(false, false, null, null, null);
+        doExpiration(emptyConfig, handler, okapiHeaders, vertx);
+      });
```
roman-barannyk, line 40:
```suggestion
      .onSuccess(tlrSettings -> doExpiration(tlrSettings, handler, okapiHeaders, vertx))
      .onFailure(t -> doExpiration(new TlrSettingsConfiguration(false, false, null, null, null), 
        handler, okapiHeaders, vertx));
```

src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
```diff
@@ -30,8 +31,18 @@ public void expireRequests(Map<String, String> okapiHeaders,
...
+  }
+
+  private void doExpiration(TlrSettingsConfiguration tlrSettings, Handler<AsyncResult<Response>> handler, Map<String, String> okapiHeaders, Vertx vertx) {
+    createRequestExpirationService(okapiHeaders, vertx, tlrSettings)
```
roman-barannyk, line 44:
```suggestion
  private void doExpiration(TlrSettingsConfiguration tlrSettings,
    Handler<AsyncResult<Response>> handler, Map<String, String> okapiHeaders, Vertx vertx) {

        createRequestExpirationService(okapiHeaders, vertx, tlrSettings)

```

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -130,6 +130,38 @@ public void canExpireASingleOpenAwaitingPickupRequest() throws InterruptedExcept
     assertThat(response.containsKey("position"), is(false));
   }
 
+  @Test
+  public void canExpireRequestWhenTlrSettingsNotReadable() throws InterruptedException,
+    MalformedURLException, TimeoutException, ExecutionException {
+    stubWithInvalidTlrSettings();
+    
+    UUID id = UUID.randomUUID();
```
roman-barannyk, line 138:
```suggestion
    MalformedURLException, TimeoutException, ExecutionException {
    
    stubWithInvalidTlrSettings();    
    UUID id = UUID.randomUUID();
```

roman-barannyk:
> 



> The failed lint tests are in the master branch, and seem to be related to the last merge. Unsure what to do about this.

It happens after merging this PR https://github.com/folio-org/mod-circulation-storage/pull/399, but there is still a chance to proceed with the merge even if the pi-schema-lint is failing

roman-barannyk (approved):

src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
```diff
@@ -30,10 +30,18 @@ public void expireRequests(Map<String, String> okapiHeaders,
 
     Vertx vertx = context.owner();
     new ConfigurationClient(vertx, okapiHeaders).getTlrSettings()
-    .compose(tlrSettings -> createRequestExpirationService(okapiHeaders, vertx, tlrSettings)
-        .doRequestExpiration())
-    .onSuccess(x -> handler.handle(succeededFuture(respond204())))
-    .onFailure(e -> handler.handle(succeededFuture(respond500WithTextPlain(e.getMessage()))));
+      .onSuccess(tlrSettings -> doExpiration(tlrSettings, handler, okapiHeaders, vertx))
+      .onFailure(t -> doExpiration(new TlrSettingsConfiguration(false, false, null, null, null),
```
OleksandrVidinieiev, line 34: This handler will user default settings in case of any error. And I think we should only do that if we receive an empty but successful response from mod-configuration (i.e. when settings are not present in its DB). If the settings we fetched are invalid, or if okapi returns an HTTP-error, or in any other case where we fail to obtain a valid set of TLR settings, we should fail as usual.

Maybe we should try and handle missing settings on `ConfigurationClient`'s side? By adding a new method, something like this:
```
public Future<TlrSettingsConfiguration> getTlrSettingsOrDefault() {
   return getTlrSettings().recover(throwable -> {
        if (settings were not found) {
          return default settings
        } else {
          rethrow throwable 
      }
     );
```

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
```diff
@@ -130,6 +130,39 @@ public void canExpireASingleOpenAwaitingPickupRequest() throws InterruptedExcept
     assertThat(response.containsKey("position"), is(false));
   }
 
+  @Test
+  public void canExpireRequestWhenTlrSettingsNotReadable() throws InterruptedException,
+    MalformedURLException, TimeoutException, ExecutionException {
+    
+    stubWithInvalidTlrSettings();
```
OleksandrVidinieiev, line 137: We will probably need more tests for different responses from mod-configuration

src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
K-Felk: As I noted in my comments above, I tried that solution and found it not viable.  Other processes in this module expect to receive errors if the TLR settings are empty-there are tests for that condition in the automated tests.  So it needs to continue to issue an empty object.

HTTP errors are checked for and caught elsewhere in this chain.  This condition only triggers if the call to the configuration settings returns a 200 but the object itself is either empty or otherwise unreadable.

src/test/java/org/folio/rest/api/RequestExpirationApiTest.java
K-Felk: Again, this condition only triggers when the TLR object itself is unreadable, either because the settings are invalid or it's empty.

src/main/java/org/folio/rest/impl/RequestExpiryImpl.java
OleksandrVidinieiev: > Other processes in this module expect to receive an empty object if the settings have not been initialized
I see that the only other process fetching TLR-configuration is TLR-feature-toggle job, and it's expecting an error from `ConfigurationClient` in case if settings were not initialized, not an empty object. The solution I offered above will not affect this behavior, as it introduces a new method which will only be invoked by request expiration service. TLR-feature-toggle job will still call the original `ConfigurationClient#getTlrSettings`.

K-Felk: Ah, I see-I didn't get that you were proposing a completely new fetch method that only the expiry code would use.  I can try that and see if it works.

PR #408, opened 2023-06-19
CIRCSTORE-412: Request does not change status when TLR settings not set

K-Felk:
It was decided that the best way to handle this was to create a alternate method in the configuration client that would return a default settings object if the TLR settings have not been initialized.  The default method assumes TLR should be disabled.  This will allow the expiration process to run if the settings have not been set.

### DIFF
--- a/src/main/java/org/folio/rest/client/ConfigurationClient.java
+++ b/src/main/java/org/folio/rest/client/ConfigurationClient.java
@@ -79,8 +79,7 @@ public class ConfigurationClient extends OkapiClient {
           log.error(errorMessage);
           return failedFuture(new HttpException(GET, url, response));
         } else {
-          String body = response.bodyAsString();
-          return succeededFuture(body);
+          return succeededFuture(response.bodyAsString());
         }
     });
   }
--- a/src/main/java/org/folio/rest/client/ConfigurationClient.java
+++ b/src/main/java/org/folio/rest/client/ConfigurationClient.java
@@ -68,6 +68,7 @@ public class ConfigurationClient extends OkapiClient {
             return failedFuture(e);
           }
         }
-      });
-    }
+      }
+    );
   }
+}
--- a/src/main/java/org/folio/rest/client/ConfigurationClient.java
+++ b/src/main/java/org/folio/rest/client/ConfigurationClient.java
@@ -27,7 +27,7 @@ public class ConfigurationClient extends OkapiClient {
   private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
   private static final String CONFIGURATIONS_URL = "/configurations/entries?query=%s";
   private static final String TLR_SETTINGS_QUERY = "module==\"SETTINGS\" and configName==\"TLR\"";
-  private static final String url = format(CONFIGURATIONS_URL, StringUtil.urlEncode(TLR_SETTINGS_QUERY));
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/client/ConfigurationClient.java
```diff
@@ -25,14 +26,55 @@ public class ConfigurationClient extends OkapiClient {
...
+
+  private Future<TlrSettingsConfiguration> parseSettings(KvConfigurations settings) {
+      return settings
+        .getConfigs().stream()
+        .findFirst()
+        .map(Config::getValue)
+        .map(JsonObject::new)
+        .map(TlrSettingsConfiguration::from)
+        .map(Future::succeededFuture)
+        .orElseGet(() -> failedFuture("Failed to find TLR configuration"));
```
OleksandrVidinieiev, line 74: Actually, I think the whole difference between the two cases boils down to the return value of this line. If TLR-configuration was fethed successfully, but the response is empty, this line should return:
- a failed future if we don't want the default value
- a succeeded future with the default value otherwise
In this case we can extract this supplier into a method parameter, and leave the rest of the code intact. Do you think this could work?

```
public Future<TlrSettingsConfiguration> getTlrSettingsOrDefault() {
    return getTlrSettings(
      () -> succeededFuture(new TlrSettingsConfiguration(false, false, null, null, null)));
  }

  public Future<TlrSettingsConfiguration> getTlrSettings() {
    return getTlrSettings(() -> failedFuture("Failed to find TLR configuration"));
  }

  private Future<TlrSettingsConfiguration> getTlrSettings(
    Supplier<Future<TlrSettingsConfiguration>> otherwise) {

    String url = format(CONFIGURATIONS_URL, StringUtil.urlEncode(TLR_SETTINGS_QUERY));

    return okapiGet(url)
      .compose(response -> {
        int responseStatus = response.statusCode();
        if (responseStatus != 200) {
          String errorMessage = format("Failed to find TLR configuration. Response: %d %s",
            responseStatus, response.bodyAsString());
          log.error(errorMessage);
          return failedFuture(new HttpException(GET, url, response));
        } else {
          try {
            return objectMapper.readValue(response.bodyAsString(), KvConfigurations.class)
              .getConfigs().stream()
              .findFirst()
              .map(Config::getValue)
              .map(JsonObject::new)
              .map(TlrSettingsConfiguration::from)
              .map(Future::succeededFuture)
              .orElseGet(otherwise);
          } catch (JsonProcessingException e) {
            log.error("Failed to parse response: {}", response.bodyAsString());
            return failedFuture(e);
          }
        }
      });
  }
```

K-Felk: I'll plug it in and test it.

src/main/java/org/folio/rest/client/ConfigurationClient.java
```diff
@@ -50,12 +62,12 @@ public Future<TlrSettingsConfiguration> getTlrSettings() {
...
           } catch (JsonProcessingException e) {
             log.error("Failed to parse response: {}", response.bodyAsString());
             return failedFuture(e);
           }
         }
       });
-  }
-}
+    }
+  }
```
OleksandrVidinieiev, line 73: The indentation was correct

OleksandrVidinieiev (approved):

PR #417, opened 2023-08-03
CIRC-1778 Actual cost records with "Open" status become "Expired"

MagzhanArtykov:
Covers [CIRC-1778](https://issues.folio.org/browse/CIRC-1778)

**Purpose**
Goal of this PR is to allow creation of Actual cost records with no Expiration date if in "Lost Item policy" option "For lost items not charged a fee/fine, close the loan after" not set.

**Approach**
Make expirationDate field in ActualCostRecord nullable

### DIFF
--- a/ramls/actual-cost-record.json
+++ b/ramls/actual-cost-record.json
@@ -325,7 +325,6 @@
   "required": [
     "lossType",
     "lossDate",
-    "expirationDate",
     "user",
     "loan",
     "item",


ramls/actual-cost-record.json
```diff
@@ -325,7 +325,6 @@
   "required": [
     "lossType",
     "lossDate",
-    "expirationDate",
```
roman-barannyk, line 328: it seems like the interface version e interface version should also be updated

PR #420, opened 2023-08-16
CIRCSTORE-426 Update request policies if ServicePoint is updated

roman-barannyk:
## Purpose
- Create a handler for handling "service point update" events. If the service point's pickupLocation field was changed from Yes (true) to No (false), it should find all of the request policies that reference the updated service point and remove it from there

Resolves: [CIRCSTORE-426](https://issues.folio.org/browse/CIRCSTORE-426)

## Approach
- add logic for removing allowed SPs from request policies if pickup location is changed from true to false;
- add tests;


src/main/java/org/folio/service/event/handler/processor/ServicePointUpdateProcessorForRequestPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+        .setJSONB(true)
+        .setVal(String.format("[\"%s\"]", servicePointId)))
+      .toList();
+    GroupedCriterias groupedCriterias = new GroupedCriterias();
+    criteriaList.forEach(criteria -> groupedCriterias.addCriteria(criteria, "OR"));
+
+    return policyRepository.get(new Criterion().addGroupOfCriterias(groupedCriterias));
+  }
+
+  private Future<List<RequestPolicy>> removeAllowedServicePoints(RequestPolicyRepository policyRepository,
```
alexanderkurash, line 94: Remove

src/main/java/org/folio/service/event/handler/processor/ServicePointUpdateProcessorForRequestPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+      log.info("removeServicePointId:: removing allowed servicePointId from Page");
+      pageAllowedServicePoints.remove(servicePointId);
+    }
+
+    Set<String> recallAllowedServicePoints = allowedServicePoints.getRecall();
+    if (recallAllowedServicePoints != null && !recallAllowedServicePoints.isEmpty()) {
+      log.info("removeServicePointId:: removing allowed servicePointId from Recall");
+      recallAllowedServicePoints.remove(servicePointId);
+    }
+    log.info("removeServicePointId:: result: {}", policy);
```
alexanderkurash, line 131: Delete request type when 0 service points left
Delete allowedServicePoints when 0 types left

src/main/java/org/folio/service/event/handler/processor/ServicePointUpdateProcessorForRequestPolicy.java
roman-barannyk: done

src/main/java/org/folio/service/event/handler/processor/ServicePointUpdateProcessorForRequestPolicy.java
roman-barannyk: done

src/main/java/org/folio/service/event/handler/processor/ServicePointUpdateProcessorForRequestPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+  protected List<Change<RequestPolicy>> collectRelevantChanges(JsonObject payload) {
+    JsonObject oldObject = payload.getJsonObject("old");
+    JsonObject newObject = payload.getJsonObject("new");
+    List<Change<RequestPolicy>> changes = new ArrayList<>();
+    String updatedServicePointId = oldObject.getString("id");
+
+    Boolean isOldServicePointPickupLocation = oldObject.getBoolean(SERVICE_POINT_PICKUP_LOCATION);
+    Boolean isNewServicePointPickupLocation = newObject.getBoolean(SERVICE_POINT_PICKUP_LOCATION);
+    if (isOldServicePointPickupLocation != null && isNewServicePointPickupLocation != null &&
+      !isNewServicePointPickupLocation && isOldServicePointPickupLocation) {
```
OleksandrVidinieiev, line 38:
```suggestion
    if (isOldServicePointPickupLocation == TRUE && isNewServicePointPickupLocation == FALSE) {
```

OleksandrVidinieiev: Those are `Boolean.TRUE` and `Boolean.FALSE`

PR #421, opened 2023-08-17
CIRCSTORE-425 Update request policies when service point is deleted

alexanderkurash:
[CIRCSTORE-425](https://issues.folio.org/browse/CIRCSTORE-425)

## Purpose
When service point is deleted, it should also be deleted from all of the request policies where it's referenced in the allowedServicePoint field.

## Approach
Handle domain event produced during service point deletion by mod-inventory-storage

### DIFF
--- a/src/main/java/org/folio/service/event/handler/ServicePointDeleteEventHandler.java
+++ b/src/main/java/org/folio/service/event/handler/ServicePointDeleteEventHandler.java
@@ -4,6 +4,7 @@ import static org.folio.kafka.KafkaHeaderUtils.kafkaHeadersToMap;
 
 import org.apache.commons.collections4.map.CaseInsensitiveMap;
 import org.folio.kafka.AsyncRecordHandler;
+import org.folio.service.event.handler.processor.ServicePointDeleteProcessorForRequestPolicy;
 
 import io.vertx.core.Context;
 import io.vertx.core.Future;
--- a/src/main/java/org/folio/service/event/handler/ServicePointDeleteEventHandler.java
+++ b/src/main/java/org/folio/service/event/handler/ServicePointDeleteEventHandler.java
@@ -10,10 +10,10 @@ import io.vertx.core.Future;
 import io.vertx.core.json.JsonObject;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
 
-public class ItemUpdateEventHandler implements AsyncRecordHandler<String, String> {
+public class ServicePointDeleteEventHandler implements AsyncRecordHandler<String, String> {
   private final Context context;
 
-  public ItemUpdateEventHandler(Context context) {
+  public ServicePointDeleteEventHandler(Context context) {
     this.context = context;
   }
 
@@ -22,10 +22,10 @@ public class ItemUpdateEventHandler implements AsyncRecordHandler<String, String
 
     JsonObject payload = new JsonObject(kafkaConsumerRecord.value());
 
-    ItemUpdateProcessorForRequest itemUpdateProcessorForRequest =
-      new ItemUpdateProcessorForRequest(context);
+    ServicePointDeleteProcessorForRequestPolicy servicePointDeleteProcessorForRequestPolicy =
+      new ServicePointDeleteProcessorForRequestPolicy(context);
 
-    return itemUpdateProcessorForRequest.run(kafkaConsumerRecord.key(),
+    return servicePointDeleteProcessorForRequestPolicy.run(kafkaConsumerRecord.key(),
       new CaseInsensitiveMap<>(kafkaHeadersToMap(kafkaConsumerRecord.headers())), payload);
   }
 }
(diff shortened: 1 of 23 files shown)


src/main/java/org/folio/service/event/handler/processor/EventProcessor.java
```diff
@@ -11,74 +10,64 @@
...
-    KafkaConsumerRecord<String, String> event) {
+  protected abstract List<Change<T>> collectRelevantChanges(JsonObject payload);
 
-    return new CaseInsensitiveMap<>(kafkaHeadersToMap(event.headers()));
-  }
+  protected abstract Future<List<T>> applyChanges(List<Change<T>> changes,
+    CaseInsensitiveMap<String, String> headers, JsonObject payload);
 
   protected <R> Future<List<R>> applyDbUpdates(List<R> objects, Collection<Change<R>> changes,
     AbstractRepository<R> repository) {
```
OleksandrVidinieiev, line 73: Repository will always be the same, should we inject it via constructor?

src/main/java/org/folio/service/event/handler/processor/ServicePointDeleteProcessorForRequestPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+      log.info("removeAllowedServicePoint:: request policy ID={}: 0 allowed service point for {} " +
+        "type, removing it from the policy", requestPolicy::getId, () -> requestType);
+
+      setAllowedServicePointsConsumer.accept(null);
+      requestPolicy.getRequestTypes().remove(requestType);
+    }
+  }
+
+  @Override
+  protected Future<List<RequestPolicy>> applyChanges(List<Change<RequestPolicy>> changes,
```
OleksandrVidinieiev, line 84: This method looks very similar in all processors:
- extract old object from payload
- instantiate repository
- fetch old object by id
- apply db updates

I think if we inject repository via constructor and extract 3d step (fetch objects by id) into an abstract method, we can make `applyChanges()` non-abstract and move it into superclass. This way you won't have to pull headers along  all the way.

src/main/java/org/folio/service/event/handler/processor/EventProcessor.java
```diff
@@ -11,74 +10,64 @@
...
-  protected UpdateEventAbstractHandler(InventoryEventType supportedEventType) {
+  protected EventProcessor(InventoryEventType supportedEventType) {
     this.supportedEventType = supportedEventType;
   }
 
-  @Override
-  public Future<String> handle(KafkaConsumerRecord<String, String> event) {
-    final String eventKey = event.key();
-    log.info("handle:: received event {}", eventKey);
+  public Future<String> run(String eventKey, CaseInsensitiveMap<String, String> headers,
```
OleksandrVidinieiev, line 29: Since all of the parameters of this method come from an instance of `KafkaConsumerRecord`, can we replace them with one parameter of type `KafkaConsumerRecord`?

src/main/java/org/folio/service/event/handler/processor/EventProcessor.java
```diff
@@ -11,74 +10,64 @@
...
+
+    return processEvent(headers, payload)
       .onSuccess(r -> log.info("handle:: event {} processed successfully", eventKey))
       .onFailure(t -> log.error("handle:: failed to process event", t))
       .map(eventKey);
   }
 
-  protected Future<List<T>> processEvent(KafkaConsumerRecord<String, String> event) {
-    JsonObject payload = new JsonObject(event.value());
+  private Future<List<T>> processEvent(CaseInsensitiveMap<String, String> headers,
```
OleksandrVidinieiev, line 40: Since all of the parameters of this method come from an instance of KafkaConsumerRecord, can we replace them with one parameter of type KafkaConsumerRecord?

src/main/java/org/folio/service/event/handler/ServicePointDeleteEventHandler.java
```diff
@@ -0,0 +1,32 @@
...
+public class ServicePointDeleteEventHandler implements AsyncRecordHandler<String, String> {
+  private final Context context;
+
+  public ServicePointDeleteEventHandler(Context context) {
+    this.context = context;
+  }
+
+  @Override
+  public Future<String> handle(KafkaConsumerRecord<String, String> kafkaConsumerRecord) {
+
```
roman-barannyk, line 23: redundant empty line

src/test/java/org/folio/rest/support/builders/JsonBuilder.java
```diff
@@ -50,6 +51,12 @@ protected void put(JsonObject representation, String property, JsonObject value)
     }
   }
 
+  protected void put(JsonObject representation, String property, JsonArray value) {
+    if(value != null) {
```
OleksandrVidinieiev, line 55:
```suggestion
    if (value != null) {
```

OleksandrVidinieiev (approved):

roman-barannyk (approved):

PR #438, opened 2023-11-10
CIRCSTORE-470: Publish circulation rules update events

OleksandrVidinieiev:
[CIRCSTORE-470](https://issues.folio.org/browse/CIRCSTORE-470)

### DIFF
--- a/src/main/java/org/folio/service/event/EntityChangedEventPublisherFactory.java
+++ b/src/main/java/org/folio/service/event/EntityChangedEventPublisherFactory.java
@@ -1,16 +1,19 @@
 package org.folio.service.event;
 
+import static org.folio.rest.tools.utils.TenantTool.tenantId;
 import static org.folio.support.kafka.topic.CirculationStorageKafkaTopic.CHECK_IN;
 import static org.folio.support.kafka.topic.CirculationStorageKafkaTopic.LOAN;
 import static org.folio.support.kafka.topic.CirculationStorageKafkaTopic.REQUEST;
-import static org.folio.rest.tools.utils.TenantTool.tenantId;
+import static org.folio.support.kafka.topic.CirculationStorageKafkaTopic.RULES;
 
 import java.util.Map;
 
 import org.folio.persist.CheckInRepository;
+import org.folio.persist.CirculationRulesRepository;
 import org.folio.persist.LoanRepository;
 import org.folio.persist.RequestRepository;
 import org.folio.rest.jaxrs.model.CheckIn;
+import org.folio.rest.jaxrs.model.CirculationRules;
 import org.folio.rest.jaxrs.model.Loan;
 import org.folio.rest.jaxrs.model.Request;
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/service/event/EntityChangedEventPublisherFactory.java
```diff
@@ -56,4 +59,15 @@ public static EntityChangedEventPublisher<String, CheckIn> checkInEventPublisher
         new CheckInRepository(vertxContext, okapiHeaders));
   }
 
+  public static EntityChangedEventPublisher<String, CirculationRules> circulationRulesEventPublisher(
+    Context vertxContext, Map<String, String> okapiHeaders) {
+
```
alexanderkurash, line 64: Add logging here

src/main/java/org/folio/support/kafka/topic/CirculationStorageKafkaTopic.java
```diff
@@ -5,7 +5,8 @@
 public enum CirculationStorageKafkaTopic implements KafkaTopic {
   REQUEST("request", 10),
   LOAN("loan", 10),
-  CHECK_IN("check-in", 10);
+  CHECK_IN("check-in", 10),
+  RULES("rules", 10);
```
alexanderkurash, line 9: Let's call them circulation-rules, considering Kafka topics being a system-wide thing

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
```diff
@@ -116,6 +118,11 @@ public static void assertUpdateEventForRequest(JsonObject oldRequest, JsonObject
     assertUpdateEvent(getLastRequestEvent(requestId), oldRequest, newRequest);
   }
 
+  public static void assertUpdateEventForCirculationRules(JsonObject oldRules, JsonObject newRules) {
+    await().until(() -> getCirculationRulesEvents().size(), greaterThan(0));
```
alexanderkurash, line 122: Should we limit the waiting time to avoid tests running forever if something goes wrong?

OleksandrVidinieiev: `await` awaits for 5 seconds by default

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
```diff
@@ -80,6 +87,13 @@ public static Collection<KafkaConsumerRecord<String, JsonObject> > getCheckInEve
     return checkInEvents.getOrDefault(checkInId, emptyList());
   }
 
+  public static Collection<KafkaConsumerRecord<String, JsonObject> > getCirculationRulesEvents() {
```
roman-barannyk, line 90:
```suggestion
  public static Collection<KafkaConsumerRecord<String, JsonObject>> getCirculationRulesEvents() {
```

src/main/java/org/folio/service/event/EntityChangedEventPublisherFactory.java
```diff
@@ -56,4 +61,17 @@ public static EntityChangedEventPublisher<String, CheckIn> checkInEventPublisher
...
+    log.info("circulationRulesEventPublisher:: creating circulation rules event publisher");
+
+    return new EntityChangedEventPublisher<>(okapiHeaders, CirculationRules::getId, NULL_ID,
+      new EntityChangedEventFactory<>(),
+      new DomainEventPublisher<>(vertxContext,
+        RULES.fullTopicName(tenantId(okapiHeaders)),
+        FailureHandler.noOperation()),
+      new CirculationRulesRepository(vertxContext, okapiHeaders));
```
roman-barannyk, line 74:
```suggestion
      new EntityChangedEventFactory<>(), new DomainEventPublisher<>(vertxContext,
        RULES.fullTopicName(tenantId(okapiHeaders)), FailureHandler.noOperation()),
      new CirculationRulesRepository(vertxContext, okapiHeaders));
```

OleksandrVidinieiev: All initializations in this class are done in this style. And I think in this case it actually makes things a bit more readable.

roman-barannyk (approved):

PR #464, opened 2024-05-31
Circstore 506

SreejaMangarapu:
Purpose: https://folio-org.atlassian.net/browse/CIRCSTORE-506 
Approach: added isDcb flag in loan.json and added few testcases for loan creation when is isDcb feild is present with values true or false

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -93,6 +93,10 @@
       "description": "Indicates whether or not this loan had its due date modified by a recall on the loaned item",
       "type": "boolean"
     },
+    "isDcb": {
+      "description": "Indicates whether or not this loan is associated for DCB use case",
+      "type": "boolean"
+    },
     "declaredLostDate" : {
       "description": "Date and time the item was declared lost during this loan",
       "type": "string",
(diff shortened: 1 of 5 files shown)


ramls/loan.json
```diff
@@ -93,6 +93,10 @@
       "description": "Indicates whether or not this loan had its due date modified by a recall on the loaned item",
       "type": "boolean"
     },
+    "isDcb": {
```
OleksandrVidinieiev, line 96: With this change we are altering the corresponding interface, which means we should also increase its version in both RAML and ModuleDescriptor

src/test/java/org/folio/rest/api/LoansApiTest.java
```diff
@@ -178,6 +178,212 @@ public void canCreateALoan()
...
+      .withOverdueFinePolicyId(overdueFinePolicyId)
+      .withLostItemPolicyId(lostItemPolicyId)
+      .withClaimedReturnedDate(claimedReturnedDate)
+      .withAgedToLostDelayedBilling(false, dateLostItemShouldBeBilled, agedToLostDate)
+      .withIsDcb(true)
+      .create();
+
+    JsonObject loan = loansClient.create(loanRequest).getJson();
+
+    assertThat("id does not match",
```
OleksandrVidinieiev, line 222: Do we really have to check every property of the loan in these test cases? We are only adding one, so maybe checking that this new field can be read and written is enough?

Vignesh-kalyanasundaram (approved):

gurleenkaurbp (approved):

OleksandrVidinieiev (approved):

PR #468, opened 2024-06-23
CIRCSTORE-496: Keep existing reference/sample data on upgrade

julianladisch:
https://folio-org.atlassian.net/browse/CIRCSTORE-496

Skip installing or resetting reference and sample data if we migrate from a version that already has installed these data.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -41,18 +41,22 @@ public class TenantRefAPI extends TenantAPI {
         .createKafkaTopics(CirculationStorageKafkaTopic.values(), tenantId))
       .compose(r -> super.loadData(attributes, tenantId, headers, vertxContext))
       .compose(r -> loadData(attributes, headers, vertxContext))
-      .compose(r -> registerModule(headers, vertxContext))
+      .compose(r -> registerModuleInPubSub(headers, vertxContext))
       .mapEmpty();
   }
 
   private Future<Integer> loadData(TenantAttributes attributes, Map<String, String> headers,
       Context vertxContext) {
 
-    log.info("Initializing of tenant's data");
-    TenantLoading tl = new TenantLoading();
-    tl.withKey(REFERENCE_KEY).withLead(REFERENCE_LEAD);
+    log.info("loadData:: Initializing tenant data");
+
+    TenantLoading tenantLoading = new TenantLoading();
+    tenantLoading.withKey(REFERENCE_KEY).withLead(REFERENCE_LEAD);
     if (isNew(attributes, "14.0.0")) {
-      tl.withIdContent()
+      log.info("loadData:: Adding reference data: loan policies, request policies, " +
+        "patron notice policies, staff slips, circulation rules, cancellation reasons");
+
+      tenantLoading.withIdContent()
         .add("loan-policy-storage/loan-policies")
         .add("request-policy-storage/request-policies")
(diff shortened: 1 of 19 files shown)


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -41,37 +40,56 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
-          .add("circulation-settings-storage/circulation-settings");
+      .compose(r -> loadData(attributes, headers, vertxContext))
+      .compose(r -> registerModule(headers, vertxContext))
+      .mapEmpty();
+  }
+
+  private Future<Integer> loadData(TenantAttributes attributes, Map<String, String> headers,
+      Context vertxContext) {
 
+    log.info("Initializing of tenant's data");
```
alexanderkurash, line 51: This logging message doesn't follow the established format

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -41,37 +40,56 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
+    TenantLoading tl = new TenantLoading();
```
alexanderkurash, line 52: tl -> tenantLoading

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -41,37 +40,56 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
-          if (res.failed()) {
-            promise.fail(res.cause());
-          } else {
-            PubSubRegistrationService.registerModule(headers, vertx)
-              .whenComplete((aBoolean, throwable) -> promise.complete());
-          }
-        });
-        return promise.future();
-      });
+  private Future<Boolean> registerModule(Map<String, String> headers, Context vertxContext) {
```
alexanderkurash, line 90: registerModuleInPubSub

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -41,37 +40,56 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
+      .mapEmpty();
+  }
+
+  private Future<Integer> loadData(TenantAttributes attributes, Map<String, String> headers,
+      Context vertxContext) {
 
+    log.info("Initializing of tenant's data");
+    TenantLoading tl = new TenantLoading();
+    tl.withKey(REFERENCE_KEY).withLead(REFERENCE_LEAD);
+    if (isNew(attributes, "14.0.0")) {
```
alexanderkurash, line 54: What should we do when we want to change reference data in newer versions? When we update this line ("14.0.0" -> new version), the reference data is going to be updated anyway. However, we'll be in control of that and can let libraries know through Release notes. Was that the intention?

src/test/java/org/folio/rest/api/TenantRefApiTests.java
```diff
@@ -456,6 +461,17 @@ public void migrationShouldNotCorrectFulfillmentPreferenceSpellingFromMigratedVe
       });
   }
 
+  @Test
+  public void keepReferenceData(TestContext context) {
```
alexanderkurash, line 465: Doesn't it make sense to test all of the reference data types with corresponding module versions, not only Cancellation reasons?

roman-barannyk (approved):

OleksandrVidinieiev (approved):

PR #489, opened 2024-09-11
[CIRCSTORE-526] Add ILR support for publishing batch request update events

roman-barannyk:
Resolves: [CIRCSTORE-526](https://folio-org.atlassian.net/browse/CIRCSTORE-526)


src/main/java/org/folio/service/request/RequestBatchResourceService.java
```diff
@@ -85,24 +85,29 @@ public void executeRequestBatchUpdate(List<Request> requests,
...
+    log.info("executeRequestBatchUpdate:: instanceId: {}, itemId: {}, requestLevel: {}, " +
+        "requests: {}", payload.getInstanceId(), payload.getItemId(), payload.getRequestLevel(),
+      payload.getRequestIds());
 
     batchResourceService.executeBatchUpdate(allDatabaseOperations, onFinishHandler)
       .compose(v -> eventPublisher.publishCreated(payload.getInstanceId(), payload));
   }
 
   private RequestQueueReordering mapRequestsToPayload(List<Request> requests) {
+    var firstRequest = requests.get(0);
```
AntonAntonich, line 101:
```suggestion
    var firstRequest = requests.iterator().next();
```

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
```diff
@@ -245,13 +247,15 @@ public void shouldPublishKafkaEventWhenUpdateRequestPositionsInBatchForTheInstan
...
     assertThat(requestsSorted.get(1).getInteger("position"), is(2));
-    assertThat(requestsSorted.get(0).getString("id"), is(secondRequest.getString("id")));
-    assertThat(requestsSorted.get(1).getString("id"), is(firstRequest.getString("id")));
+    assertThat(requestsSorted.get(0).getString("id"), is(secondRequestId));
+    assertThat(requestsSorted.get(1).getString("id"), is(firstRequestId));
 
-    assertRequestQueueReorderingEvent(instanceId.toString(), List.of(
-      firstRequest.getString("id"), secondRequest.getString("id")));
+    assertRequestQueueReorderingEvent(instanceId.toString(),
+      requestsSorted.get(1).getString("itemId"), List.of(firstRequestId, secondRequestId), TITLE);
```
AntonAntonich, line 258: m.b. it's a good idea to extract constants ?

AntonAntonich (approved):

src/main/java/org/folio/service/request/RequestBatchResourceService.java
roman-barannyk: Since we work with List in this method (not Collection), accessing an element by index will be more efficient for array-based implementation with constant time complexity O(1)

src/test/java/org/folio/rest/api/RequestBatchAPITest.java
roman-barannyk: assertRequestQueueReorderingEvent is a universal method that is also applicable to queues with ITEM requestLevel requests

alexanderkurash (approved):

PR #491, opened 2024-09-19
CIRC-2141 Add itemLocationCode to request object schema

JanisSaldabols:


ramls/request.json
```diff
@@ -231,6 +231,10 @@
       "description": "Request fields used for search",
       "type": "object",
       "$ref": "request-search-index.json"
+    },
+    "itemLocationCode": {
```
OleksandrVidinieiev, line 235: Can we have a test which checks that a request can actually be created with this new property?

JanisSaldabols: Added test

OleksandrVidinieiev (approved):

PR #524, opened 2025-07-04
CIRCSTORE-578 use at location

nielserik:
Some libraries have materials that patrons can loan for a given period but cannot bring out of the library. Instead items must be picked up and put back on a hold shelf with each day of use. Each use must be registered to keep track of the items and for statistics. See also [UXPROD-1835](https://folio-org.atlassian.net/browse/UXPROD-1835). 

This PR adds new, optional properties to the loan policy and the loan. In the loan policy it can be specified that the items under the policy must be used at the location. The loan has new properties for registering the usage at location. 

The processing of the new properties will happen in mod-circulation, [CIRC-2308](https://folio-org.atlassian.net/browse/UXPROD-1835).

The version numbers of `loan-storage` and `loan-policy-storage` are bumped a minor due to the new, optional properties.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -62,7 +62,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "7.3",
+      "version": "7.4",
       "handlers": [
         {
           "methods": ["GET"],
@@ -165,7 +165,7 @@
     },
     {
       "id": "loan-policy-storage",
-      "version": "2.3",
+      "version": "2.4",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/loan-policy-storage.raml
+++ b/ramls/loan-policy-storage.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Loan Policy Storage
-version: v2.3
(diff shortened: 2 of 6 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -62,7 +62,7 @@
     },
     {
       "id": "loan-storage",
-      "version": "7.3",
+      "version": "7.4",
```
alexanderkurash, line 65: Are there any modules that have a dependency on loan-storage v7.3 or loan-policy-storage v2.3.

MikeTaylor:
Can I ask what's holding up the merging of this very simple PR? It's foundational to a whole tree of other things that need merging in other back-end and UI modules.

descriptors/ModuleDescriptor-template.json
nielserik: Right, there can be dependencies that we don't know of. FWIW we _do_ know that mod-circulation has a dependency on loan-storage 7.3.

MikeTaylor:
Glory hallelujah! Now on to https://github.com/folio-org/mod-circulation/pull/1603 :-)

PR #569, opened 2026-04-23
MODFQMMGR-1112 Update FQM metadata for actual cost record

mweaver-ebsco:


ramls/actual-cost-record.json
```diff
@@ -114,6 +114,10 @@
           "x-fqm-visible-by-default": true,
           "x-fqm-visibility": "all",
           "x-fqm-essential": true,
+          "x-fqm-source": {
+            "entityTypeId": "e7717b38-4ff3-4fb9-ae09-b3d0c8400710",
```
mweaver-ebsco, line 118: This is the simple_groups ID. No need to do the full source config here when we can just pull it from the entity type with that data

mweaver-ebsco:
Note: This PR targets `master` because the changes don't need to actually make it into any particular mod-circulation-storage release (the changes will end up in a trillium release for mod-fqm-manager, but that happens in a different repo). These changes have no runtime impact on mod-circulation-storage.

ncovercash (approved):

PR #570, opened 2026-04-27
CIRCSTORE-612 Handle empty CQL set (==()) in queries

roman-barannyk:
Resolves: [CIRCSTORE-612](https://folio-org.atlassian.net/browse/CIRCSTORE-612)


alexanderkurash:
Needs to be implemented in mod-circulation

